PR #1, opened 2019-07-01
Add logo

fabien0102:
image

Note: I let you update the readme with the correct link, it's just to have a proper preview for the validation (since we rely on master assets ^^)


fabien0102:
https://github.com/nicoespeon/abracadabra/tree/add-logo

Also, I'm wondering why we have this pattern for images, relative path works well and don't depend on another branch

nicoespeon:
> Also, I'm wondering why we have this pattern for images, relative path works well and don't depend on another branch

Relative paths doesn't work everywhere. Typical example is yarn / npm website: if path to the image was relative in the README, the image won't appear on the package manager.

I'm not sure about VS Code Marketplace yet, but I got used to refer to the online hosted version because of that ;-)

nicoespeon:
Also: awesooooooooooome logo! Thanks buddy!

nicoespeon:
@fabien0102 here's the answer for VS Code, trying to create a package:

image

So we could have relative paths & provide the repo URL as an argument. But for now I prefer to use the hosted URL to make it work anywhere by default (as long as you're connected).

nicoespeon:
I put details of that decision in an ADR, so it's explicit: https://github.com/nicoespeon/abracadabra/blob/master/docs/adr/0003-use-hosted-image-links-in-readme.md

We can revisit that decision later. I put all the information around the decision in the file.

PR #10, opened 2019-07-14
Custom testEach() to replace Jest it.each()

### LINKED ISSUE #2: Replace Jest `it.each` with a custom implementation
Jest `it.each` seemed great, but they are limited. 

Specifically, I want to be able to `.only` a single test of the list when I'm working on it, which is not doable. I know this can be done at runtime through Jest runner, but I use other tools that are integrated with the editor (so I don't look at the terminal). Thus, I need `.only`.

nicoespeon:
Fixes #2

### DIFF
--- a/src/tests-helpers.ts
+++ b/src/tests-helpers.ts
@@ -1,0 +1,22 @@
+import { Code } from "./refactorings/editor/i-write-code";
+import { Selection } from "./refactorings/editor/selection";
+
+export { testEach };
+
+function testEach<T>(
+  label: string,
+  assertions: (Assertion & T)[],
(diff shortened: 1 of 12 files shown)


src/tests-helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+export { testEach };
+
+function testEach<T>(
+  label: string,
+  assertions: (Assertion & T)[],
+  fn: (...args: (Assertion & T)[]) => any
+) {
+  describe(label, () => {
+    assertions.forEach(assertion => {
+      const test = assertion.only ? it.only : it;
```
nicoespeon: ☝️ 😁

PR #22, opened 2019-08-14
Rewrite "Extract Variable"

nicoespeon:
"Extract Variable" was one of the first refactorings I implemented. When I did it, I didn't transformed the AST. Instead, I got information from it, then asked VS Code to read the code and manipulated the strings.

While it was working, it bothered me for few reasons:
1. Inconsistency with the other refactorings, since the approach was different
1. Custom code in Selection that was only used for this use case, because of the approach
1. Necessity to have and maintain a `ReadThenWrite` adapter (there are still 2 refactorings using it)
1. Make it harder to combine with other refactorings because it doesn't depend on the same adapter 
1. Because we were counting on strings, edge cases implied adding more custom code that couldn't be re-used (e.g. JSX flags for adding `{` around `extracted`). Using an appropriate model (👋 AST) should limit that.

Each of these reasons could have been solved with other techniques, but rewriting the refactoring to use the same approach solved them all.

All the 68 tests I wrote to cover this feature still pass. If this introduces a regression, it would be on patterns I didn't test before.

## What changes

As the extracted code is a new node in the AST, extracted objects / arrays will be formatted. Whereas before, it was the exact text string that was cut & pasted. 

```js
console.log({ firstName: "Jane", lastName: "Doe" });
```

Extraction, before this PR:

```js
const extracted = { firstName: "Jane", lastName: "Doe" };
console.log(extracted);
```

Extraction, after this PR:

```js
const extracted = {
  firstName: "Jane", 
  lastName: "Doe" 
};

console.log(extracted);
```

Before was better. But I'm OK to accept the tradeoff for the moment and see how it goes (my code is usually formatted on save/commit, so I'm fine). 

If that cause you troubles, I count on you to tell me ;-) 

## Remaining work

- [ ] Set cursor position after rewrite to be on the new variable identifier (should fix issue with JSX extractions that didn't get renamed because of `{`)
- [ ] Update Changelog

### DIFF
--- a/src/refactorings/extract-variable/extract-variable.test.ts
+++ b/src/refactorings/extract-variable/extract-variable.test.ts
@@ -95,40 +95,35 @@ console.log(extracted);`);
         code: `console.log("Hello!");`,
         selection: Selection.cursorAt(0, 12),
         expected: `const extracted = "Hello!";
-console.log(extracted);`,
-        skip: true
+console.log(extracted);`
       },
       {
         description: "a number",
         code: `console.log(12.5);`,
         selection: Selection.cursorAt(0, 12),
         expected: `const extracted = 12.5;
-console.log(extracted);`,
-        skip: true
+console.log(extracted);`
       },
       {
         description: "a boolean",
         code: `console.log(false);`,
         selection: Selection.cursorAt(0, 12),
         expected: `const extracted = false;
-console.log(extracted);`,
-        skip: true
+console.log(extracted);`
       },
       {
         description: "null",
         code: `console.log(null);`,
         selection: Selection.cursorAt(0, 12),
         expected: `const extracted = null;
-console.log(extracted);`,
-        skip: true
+console.log(extracted);`
       },
       {
         description: "undefined",
         code: `console.log(undefined);`,
         selection: Selection.cursorAt(0, 12),
         expected: `const extracted = undefined;
-console.log(extracted);`,
-        skip: true
+console.log(extracted);`
       },
       {
         description: "an array",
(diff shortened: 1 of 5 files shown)


src/refactorings/extract-variable/extract-variable.test.ts
```diff
@@ -140,18 +140,19 @@ console.log(extracted);`
...
         description: "an object",
         code: `console.log({ one: 1, foo: true, hello: 'World!' });`,
         selection: Selection.cursorAt(0, 12),
-        expected: `const extracted = { one: 1, foo: true, hello: 'World!' };
+        expected: `const extracted = {
+  one: 1,
+  foo: true,
+  hello: 'World!'
+};
+
```
nicoespeon: Examples of patterns that gets formatted.

nicoespeon:
So, I thought about it and I don't think loosing the formatting is an acceptable tradeoff in the end. This is likely a feature that would be asked and the way to do it *is* to read the actual code then write it. For this refactoring, I think it makes sense… even if it's more complex.

I tackled some issues I listed with other fixes:

> * Custom code in Selection that was only used for this use case, because of the approach

This code belonged to AST. [I moved and simplified it](https://github.com/nicoespeon/abracadabra/commit/4dddf27c95abcafd787a05972e65c67bea7b4aae).

> * Make it harder to combine with other refactorings because it doesn't depend on the same adapter

So this was a [data clump smell](https://refactoring.guru/smells/data-clumps). I kept passing `write` and `showErrorMessage` around in recent refactorings. Some would take `readThenWrite` instead, or `delegateToEditor`.

I extracted a class to merge these concepts together. Turns out this is the "Editor" concept.

I didn't want to have one big "Editor" interface at first as I didn't want to have a god class. But from usage, it becomes clear it's a missing concept.

---

So there is the `Editor` interface, and 2 implementations: 
- `InMemoryEditor` that I use in unit tests
- `VSCodeEditor` which makes it work in integration with VS Code API

So different approach of Extract Variable will still add complexity, but I think it's the cost of preserving the extracted variable style as much as possible. If it becomes a real maintenance problem, I might revise this decision.

Issue #24, opened 2019-08-17
New Refactoring: Add Braces to If Statement


nicoespeon:
## Is this request related to a problem? Please describe.

I do use one-liner If Statements to write guard clause.

```js
function doSomething() {
  if (!isValid) return;

  // … rest of the code
}
```

But sometimes, the guard clause might evolve and do something else. In that scenario, I want to add braces.

```js
function doSomething() {
  if (!isValid) {
    // … now I can add some logic here
    return;
  }

  // … rest of the code
}
```

Adding braces can be a bit painful because the editor will likely insert the closing one. So I usually have to:
- put the cursor at the right place
- insert the opening brace
- eventually delete the closing one
- put the cursor at the right place again
- insert the closing brace

## Describe the solution you'd like

Some extensions might help me wrap selection into braces. I want almost the same thing, but something simpler and specific to If Statements: a quick fix that would trigger whenever the selection is inside the correct node.

Given:

```js
function doSomething() {
  if (!isValid) return;

  console.log("Something");
}
```

When I trigger the refactoring on the If Statement node,

Then I want:

```js
function doSomething() {
  if (!isValid) {
    return;
  }

  console.log("Something");
}
```

## Additional context

It's a similar refactoring than _[Add Braces to Arrow Function](https://github.com/nicoespeon/abracadabra/tree/master/src/refactorings/add-braces-to-arrow-function)_. 

You can use `yarn new` to help you bootstrap the refactoring.

I'm here to help.

Example on Webstorm:

![image](https://user-images.githubusercontent.com/1094774/63211867-0ec20080-c0cb-11e9-8e0d-c2fbdc62209c.png)

PR #54, opened 2019-12-04
Issue #24: Add braces to if statement

HEYGUL:
#24 implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "onCommand:abracadabra.replaceBinaryWithAssignment",
     "onCommand:abracadabra.splitDeclarationAndInitialization",
     "onCommand:abracadabra.splitIfStatement",
+    "onCommand:abracadabra.addBracesToIfStatement",
     "onLanguage:javascript",
     "onLanguage:javascriptreact",
     "onLanguage:typescript",
@@ -231,6 +232,11 @@
       {
         "command": "abracadabra.splitIfStatement",
         "title": "Split If Statement",
+        "category": "Abracadabra"
+      },
+      {
+        "command": "abracadabra.addBracesToIfStatement",
+        "title": "Add Braces to If Statement",
         "category": "Abracadabra"
       }
     ],
@@ -632,6 +638,10 @@
         {
           "command": "abracadabra.splitIfStatement",
           "when": "editorLangId == typescriptreact"
+        },
+        {
+          "command": "abracadabra.addBracesToIfStatement",
+          "when": "editorLangId == javascript"
         }
       ]
     }
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
   },
   "activationEvents": [
     "onCommand:abracadabra.addBracesToArrowFunction",
+    "onCommand:abracadabra.addBracesToIfStatement",
     "onCommand:abracadabra.bubbleUpIfStatement",
     "onCommand:abracadabra.convertForToForeach",
     "onCommand:abracadabra.convertIfElseToTernary",
@@ -106,7 +107,6 @@
     "onCommand:abracadabra.replaceBinaryWithAssignment",
     "onCommand:abracadabra.splitDeclarationAndInitialization",
     "onCommand:abracadabra.splitIfStatement",
-    "onCommand:abracadabra.addBracesToIfStatement",
     "onLanguage:javascript",
     "onLanguage:javascriptreact",
     "onLanguage:typescript",
@@ -117,6 +117,11 @@
       {
         "command": "abracadabra.addBracesToArrowFunction",
         "title": "Add Braces to Arrow Function",
+        "category": "Abracadabra"
+      },
+      {
+        "command": "abracadabra.addBracesToIfStatement",
+        "title": "Add Braces to If Statement",
         "category": "Abracadabra"
       },
       {
@@ -233,11 +238,6 @@
         "command": "abracadabra.splitIfStatement",
         "title": "Split If Statement",
         "category": "Abracadabra"
-      },
-      {
-        "command": "abracadabra.addBracesToIfStatement",
-        "title": "Add Braces to If Statement",
-        "category": "Abracadabra"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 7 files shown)


HEYGUL:
I open the PR as soon as possible to take opportunity to get feedback as soon as possible during the implement of my very first `Abracadabra` feature.
Feel free to comment and/or give advice. 👍

src/refactorings/add-braces-to-if-statement/add-braces-to-if-statement.ts
```diff
@@ -0,0 +1,59 @@
...
+    return;
+  }
+
+  await editor.write(updatedCode.code);
+}
+
+function hasIfStatementToAddBraces(ast: t.AST, selection: Selection): boolean {
+  let result = false;
+
+  t.traverseAST(ast, {
```
nicoespeon: Here we can use the `createVisitor()` method to get rid of the duplicated logic.

We would replace this with 
```js
t.traverseAST(ast, createVisitor(selection, () => (result = true)));
```

src/refactorings/add-braces-to-if-statement/add-braces-to-if-statement.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  testEach<{ code: Code; selection?: Selection; expected: Code }>(
+    "should add braces to if statement",
+    [
+      {
+        description: "basic scenario",
+        code: "if (!isValid) return;",
+        selection: Selection.cursorAt(0, 17),
+        expected: `if (!isValid) {
+  return;
+}`
```
nicoespeon: 👍 That's great!

Now can we think about other use-cases? What would happen for:

```js
if (isValid) 
  doSomething();
else
  doAnotherThing();
```

Ideally, it should only add braces to the `if` statement (= consequent), so I think it already works. But then, if we trigger that from the `else` (= alternate), I'd expect to add braces on the `else` statement instead.

That might be more tricky to implement. Let me know if you need help to figure this out. 

And this one should not change (which, I think, is already the case):

```js
if (isValid) {
  doSomething();
}
```

src/refactorings/add-braces-to-if-statement/index.ts
```diff
@@ -0,0 +1,20 @@
...
+    operation: addBracesToIfStatement,
+    title: "Add Braces To If Statement"
+  },
+  actionProvider: {
+    message: "Add braces to if statement",
+    canPerform: hasIfStatementToAddBraces
+  }
+};
+
+export default config;
```
nicoespeon: All of this is very nice. The essentials of the feature seems to be here. We need to declare the refactoring in the `extension.ts` file and in the `package.json`, like the others, so it becomes available. 

You should be able to see it if you run the extension host debugger (`F5`).

src/refactorings/add-braces-to-if-statement/add-braces-to-if-statement.test.ts
HEYGUL: Yes if have a full list of use-cases. My first goal was to get a first understanding of AST manipulation and get my first green test (TDD forever).

src/refactorings/add-braces-to-if-statement/add-braces-to-if-statement.ts
HEYGUL: I agree.

src/refactorings/add-braces-to-if-statement/index.ts
HEYGUL: Thank you, I will have a look to those 2 files.

nicoespeon:
@HEYGUL do you need my help? No pressure, but I want to check in case you'd be blocked with something—that's fine if you just didn't have time to move on this, there's no hurry.

HEYGUL:
> @HEYGUL do you need my help? No pressure, but I want to check in case you'd be blocked with something—that's fine if you just didn't have time to move on this, there's no hurry.

Hey @nicoespeon, I just did not have time till yesterday. I took into account some of your feedbacks and I am working on the `else` selection scenario :-)

nicoespeon:
@HEYGUL perfect, let me know when you're done 👍

HEYGUL:
I implement both consequent and alternate selections to add braces to, but if you select the `else` or `if` keywords themselves, the `selection.isInsideNode` function returns false.
I am not sure how this refactoring shall work 🤔 @nicoespeon I need your advice and point of view about.

I updated README for this refactoring but I do not know how you generate the demo gifs, any tips are welcome.

nicoespeon:
@HEYGUL I'll have a look at this next week when I'm back to work. Thanks for contributing 😉 

As for the gifs, I can create them. But if you want to know, I'm using https://recordit.co/ and it's super convenient!

nicoespeon:
@HEYGUL your implementation was very good. It didn't miss a lot to complete the feature.

Indeed, I'd expect the UX to handle when the cursor is on the `if` or `else` keyword, so usage is more intuitive. It turns out you weren't far from solving this. In fact, instead of checking for `isInsideNode()`, we had to check the position of the start of the cursor:

- if it's before the end of the `consequent` node, then it means we're on the `if` part
- else (it's after the end of the `consequent` node), then it means we're on the `else` part

So the code becomes:

```js
      if (!t.isSelectableNode(path.node.consequent)) return;
      const endOfConsequent = Position.fromAST(path.node.consequent.loc.end);

      if (selection.start.isBefore(endOfConsequent)) {
        path.node.consequent = statementWithBraces(path.node.consequent);
        return;
      }

      if (path.node.alternate) {
        path.node.alternate = statementWithBraces(path.node.alternate);
      }
```

And the `statementWithBraces()`:

```js
function statementWithBraces(node: t.Statement): t.Statement {
  return t.isBlockStatement(node) ? node : t.blockStatement([node]);
}
```

Finally, I checked for a nasty scenario: **nested if statements**. 
With current implementation, it would always take the outer if statement and stop.

To solve this, we need to check if a children if statement would match. That's a classic check we need to do, so I think there's a pattern I could refactor. But I didn't yet. Nevermind, I implemented the fix along with tests.

IMO, we're good to go. I'll update the Changelog & README. This will probably be released by the end of the week.

Do you have any question before I merge?

package.json
```diff
@@ -632,6 +638,10 @@
         {
           "command": "abracadabra.splitIfStatement",
           "when": "editorLangId == typescriptreact"
+        },
+        {
+          "command": "abracadabra.addBracesToIfStatement",
+          "when": "editorLangId == javascript"
```
nicoespeon: Can you duplicate this command declaration to enable it for all the other 3 languages?

If you look closely, you'll notice that each refactoring is declared for the 4 languages (js, jsx, ts, tsx). That's a lot of boilerplate, but I didn't put an abstraction on top to simplify that—yet.

nicoespeon: Nevermind, it was a simple boilerplate: I did it so I can merge your PR and release this 😉

nicoespeon:
Merging this. Thanks for your contribution @HEYGUL 👍

Issue #18, opened 2019-07-25
Do you noticed performance issue? Let us know.


nicoespeon:
I think the way we provide the Quick Fixes might lead to performance issues. So far, it seems to be fine. But I'm not sure how much current implementation will scale.

**If you experienced performance issues because of the extension, please let us know here 🙏**

## Details

Today, each Quick Fix is independent. 

To tell VS Code if it should propose a refactoring as a Quick Fix, we traverse the code, check if the refactoring could be made from current selection, and tell VS Code if we can.

Example of such behavior, implemented in `action-provider.ts`:

https://github.com/nicoespeon/abracadabra/blob/28273cf7eddae39b1db8473502fbcfe2cbe6cd73/src/refactorings/convert-ternary-to-if-else/action-provider.ts#L15-L35

https://github.com/nicoespeon/abracadabra/blob/28273cf7eddae39b1db8473502fbcfe2cbe6cd73/src/refactorings/convert-ternary-to-if-else/convert-ternary-to-if-else.ts#L27-L29

The performance problem would be due to the fact that **each refactoring parses, traverses and transforms the AST**. At some point, this could be too much. 

If this is confirmed, the solution would be to traverse the AST less. Ideally, we would traverse it once and try to figure out what list of refactorings could be performed from that.

The challenge would be to provide individual action providers to VS Code. That's why current implementation is very naive, so every refactoring is independent—*which simplifies development a lot!*

---

**But until we got actual confirmation that this is a concern, we'll keep current implementation.** #makeItWork #makeItRight #makeItFast

![hack-tv](https://user-images.githubusercontent.com/1094774/61837436-c65c4d80-ae52-11e9-8fb6-7bb88d9f1729.gif)

simonhoss:
Hi. I have the case when I switch between different files. That it takes a long time until it shows the git changes (the blue marker on the left) and when I have wallaby activated it takes also some time until the rectangles on the side occurs

nicoespeon:
Hi @simonhoss 👋 

I'm not sure what you mentioned is related to the Abracadabra extension. Wallaby takes also some time to boot indeed, but that's on Wallaby 👐 

I'll run perf analysis and come up with improvements for the extension if I find any issue. Overall, on my machine, despite working on big files I find Abracadabra to still have decent perfs today. However, I can already think about some improvements that would make it more scalable (e.g. we can add more refactorings without impacting the overall extension perfs).

nicoespeon:
[...]

PR #77, opened 2020-02-14
[Draft] Performance improvement (General Idea)

visusnet:
We have experienced some grave performance issues which made the extension unusable for us. The root cause is known (https://github.com/nicoespeon/abracadabra/issues/18): Every refactoring traverses the AST.

This PR is not meant to be merged. It is merely intended as a conversation starter. @timvancleef and I have talked about a feasible approach and this is a draft which could work.

There are a couple of things that I don't like:
- Backwards compatibility is required because (at least) refactorings that use codemods cannot be refactored in this way. Codemod refactorings do not traverse the AST (... well, they do, but in another library which is outside of our reach).
- I had to delete a couple of tests and was too lazy to write new tests. Actually, there are no tests for this approach. Again: It is intended as a draft. Do not merge this PR!
- The code is far from clean.
- I have not tested whether this _really_ is worth the effort. I have tested it with large files and it felt faster... but I haven't profiled the extension. (This is great, thanks! https://www.nicoespeon.com/en/2019/11/fix-vscode-extension-performance-issue/)
- I have not tested whether I have broken some of the refactorings. I have merely smoke-tested a couple of them.

@nicoespeon What do you think? Is this approach going in the right direction? What are your thoughts?

### DIFF
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -1,9 +1,9 @@
 import * as vscode from "vscode";
 
 import { createSelectionFromVSCode } from "./editor/adapters/vscode-editor";
-import { Selection } from "./editor/selection";
 import { RefactoringWithActionProvider } from "./types";
 import * as t from "./ast";
+import { Selection } from "./editor/selection";
 
 export { RefactoringActionProvider };
 
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     const ast = t.parse(document.getText());
     const selection = createSelectionFromVSCode(range);
 
-    return this.refactorings
-      .filter(refactoring => this.canPerform(refactoring, ast, selection))
-      .map(refactoring => this.buildCodeActionFor(refactoring));
+    const applicableRefactorings: RefactoringWithActionProvider[] = [];
+    t.traverseAST(ast, {
+      enter: (path: t.NodePath<any>) => {
+        this.refactorings.forEach(
+          this.checkCanPerform(
+            path,
+            selection,
+            (
+              matchedPath: t.NodePath<any>,
+              refactoring: RefactoringWithActionProvider
+            ) => {
+              if (refactoring.actionProvider.canPerformRefactoringMutator) {
+                applicableRefactorings.push(
+                  refactoring.actionProvider.canPerformRefactoringMutator(
+                    matchedPath,
+                    refactoring
+                  )
+                );
+              } else {
+                applicableRefactorings.push(refactoring);
+              }
+            }
+          )
+        );
+        path.type;
+      }
+    });
+
+    return [
+      ...applicableRefactorings.map(refactoring =>
+        this.buildCodeActionFor(refactoring)
+      ),
+      ...this.refactorings
+        .filter(refactoring => this.canPerform(refactoring, ast, selection))
+        .map(refactoring => this.buildCodeActionFor(refactoring))
+    ];
+  }
+
+  private checkCanPerform(
+    path: any,
+    selection: any,
+    onCanPerform: (
+      matchedPath: t.NodePath<any>,
+      refactoring: RefactoringWithActionProvider
+    ) => void
+  ) {
+    return (refactoring: RefactoringWithActionProvider) => {
+      if (refactoring.actionProvider.canPerformVisitorFactory) {
+        const visitor: any = refactoring.actionProvider.canPerformVisitorFactory(
+          selection,
+          path => onCanPerform(path, refactoring),
+          refactoring
+        );
+
+        const node: t.Node = path.node;
+
+        try {
+          if (typeof visitor[node.type] === "function") {
+            visitor[node.type](path);
+          } else if (typeof visitor[node.type] === "object") {
+            visitor[node.type].enter(path);
+          }
+        } catch (_) {
+          // Silently fail, we don't care why it failed (e.g. code can't be parsed).
+        }
+      }
+    };
   }
 
   private canPerform(
@@ -32,7 +96,10 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     selection: Selection
   ) {
     try {
-      return refactoring.actionProvider.canPerform(ast, selection);
+      return (
+        typeof refactoring.actionProvider.canPerform === "function" &&
+        refactoring.actionProvider.canPerform(ast, selection)
+      );
     } catch (_) {
       // Silently fail, we don't care why it failed (e.g. code can't be parsed).
       return false;
--- a/src/refactorings/convert-if-else-to-switch/convert-if-else-to-switch.ts
+++ b/src/refactorings/convert-if-else-to-switch/convert-if-else-to-switch.ts
@@ -2,7 +2,7 @@ import { Editor, Code, ErrorReason } from "../../editor/editor";
 import { Selection } from "../../editor/selection";
 import * as t from "../../ast";
 
-export { convertIfElseToSwitch, hasIfElseToConvert };
+export { convertIfElseToSwitch, hasIfElseToConvertVisitorFactory };
 
 async function convertIfElseToSwitch(
   code: Code,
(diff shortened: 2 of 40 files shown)


nicoespeon:
Thanks @visusnet I'll have a look this week-end ;-)

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     const ast = t.parse(document.getText());
     const selection = createSelectionFromVSCode(range);
 
-    return this.refactorings
-      .filter(refactoring => this.canPerform(refactoring, ast, selection))
-      .map(refactoring => this.buildCodeActionFor(refactoring));
+    const applicableRefactorings: RefactoringWithActionProvider[] = [];
+    t.traverseAST(ast, {
```
nicoespeon: I think that make sense indeed.

We already optimized this check a bit by parsing the code into AST once. It's only logical that we'd want to traverse the AST once and gather the feedback from the different refactorings 👍 

I agree with the direction of the PR.

From experience, what could go wrong would be one refactoring having side-effect on the AST traversal (e.g. stopping the traversal or modifying the traversed tree). I think it should be fine because most checks just set a boolean, but I'm not 100% confident on this.

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+            ) => {
+              if (refactoring.actionProvider.canPerformRefactoringMutator) {
+                applicableRefactorings.push(
+                  refactoring.actionProvider.canPerformRefactoringMutator(
+                    matchedPath,
+                    refactoring
+                  )
+                );
+              } else {
+                applicableRefactorings.push(refactoring);
```
nicoespeon: I guess this is an intermediate step, until all refactorings are migrated to use the new pattern?

If so, I'd left a comment to explain "why" in the code (that could be surprising). But I think it's a good approach, so we can progressively ship this. I wouldn't like we have a long-living branch since it potentially conflict with other changes we could do.

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+              }
```
nicoespeon: Speaking about "comments to explain why", I think it would worth creating a new ADR document to explain this change too, once we take a decision 👍

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+      ),
+      ...this.refactorings
+        .filter(refactoring => this.canPerform(refactoring, ast, selection))
+        .map(refactoring => this.buildCodeActionFor(refactoring))
+    ];
+  }
+
+  private checkCanPerform(
+    path: any,
+    selection: any,
```
nicoespeon: 🍰 Probably part of the draft proposal, but we should pay attention not to left unnecessary `any`.

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+    onCanPerform: (
+      matchedPath: t.NodePath<any>,
+      refactoring: RefactoringWithActionProvider
+    ) => void
+  ) {
+    return (refactoring: RefactoringWithActionProvider) => {
```
nicoespeon: Maybe we could simplify this and not return a function from this method. I think this can be done by pulling up the `(refactoring: RefactoringWithActionProvider) => {}` back in the `forEach()` caller. What do you think?

src/action-providers.ts
```diff
@@ -21,9 +21,73 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+          path => onCanPerform(path, refactoring),
+          refactoring
+        );
+
+        const node: t.Node = path.node;
+
+        try {
+          if (typeof visitor[node.type] === "function") {
+            visitor[node.type](path);
+          } else if (typeof visitor[node.type] === "object") {
```
nicoespeon: I wonder if we could standardize this so it streamlines the code. But overall, I get the idea 👍

src/refactorings/convert-if-else-to-switch/convert-if-else-to-switch.ts
```diff
@@ -19,11 +19,11 @@ async function convertIfElseToSwitch(
...
-function hasIfElseToConvert(ast: t.AST, selection: Selection): boolean {
-  let result = false;
-  t.traverseAST(ast, createVisitor(selection, () => (result = true)));
-
-  return result;
+function hasIfElseToConvertVisitorFactory(
+  selection: Selection,
+  onMatch: (path: t.NodePath<any>) => void
+): t.Visitor {
+  return createVisitor(selection, onMatch);
```
nicoespeon: I like how it will remove a lot of duplicated code =)

nicoespeon: Since we could even simplify that further and use `createVisitor` directly.

src/refactorings/convert-ternary-to-if-else/convert-ternary-to-if-else.ts
```diff
@@ -19,32 +19,31 @@ async function convertTernaryToIfElse(
...
-        result = true;
+        onMatch(path);
       }
 
       if (
         t.isVariableDeclarator(parentPath.node) &&
         t.isVariableDeclaration(parentPath.parent)
       ) {
-        result = true;
+        onMatch(path);
```
nicoespeon: Oh I see, this one will require more thoughts indeed.

nicoespeon: Certainly due to the fact that some refactorings were not yet converted to the `createVisitor()` pattern. We should refactor that first to get these follow the same (simple) implementation.

src/refactorings/merge-if-statements/index.ts
```diff
@@ -13,11 +15,21 @@ const config: RefactoringWithActionProvider = {
   actionProvider: {
     message: "Merge if statements",
 
-    canPerform(ast: t.AST, selection: Selection) {
-      const { mergeAlternate, canMerge } = tryMergeIfStatements(ast, selection);
-      this.message = mergeAlternate ? "Merge else-if" : "Merge if statements";
+    canPerformVisitorFactory: canMergeStatementsVisitorFactory,
 
-      return canMerge;
+    canPerformRefactoringMutator(
```
nicoespeon: Not 100% sold on the name yet (not sure what "mutator" refers to, regarding the function output).

src/action-providers.ts
visusnet: Actually, no, that's not what I thought. Some refactorings change how the action provider is displayed. Such refactorings can implement canPerformRefactoringMutator which mutates the refactoring object (which contains the actionProvider object), thus is able to mutate the action provider message. Refactorings that do not implement this function will just be added without any mutation happening.

visusnet: But I agree... It is not very easy to understand. Maybe, just maybe, refactorings could be a class that extends an abstract refactoring class which implements a no-op mutator. But usually, I don't like to use inheritance to obscure a simple condition. What do you think?

src/action-providers.ts
visusnet: True. I will change that in the next proposal.

src/action-providers.ts
visusnet: Yes, that works. It will be part of the next proposal.

src/action-providers.ts
visusnet: Unfortunately, visitors can either be an object or a function. It would be confusing for devs to not be able to use one of those even though babel's AST traversal function explicitly can do both. However, I think this can be improved by extracting a function that encapsulates both variants.

src/action-providers.ts
visusnet: I agree. It should be documented that .stop() is not allowed in such refactoring implementations.

Issue #79, opened 2020-02-21
Performance improvements


visusnet:
This is a continuation of #77.

nicoespeon:
Nice, thank you @visusnet 🙏 

I have a busy week (meetups & conference all week long 😅) so I'll review all of that next week-end.

![canada-drink](https://user-images.githubusercontent.com/1094774/75120520-69c84780-565a-11ea-9ad7-6acece08fb67.gif)

Have a nice week!

nicoespeon:
Triggering Abracadabra on a 2.3k LOC file.

Before:

![image](https://user-images.githubusercontent.com/1094774/77814031-e26c5a80-7083-11ea-9d67-13e9cad89423.png)

After:

![image](https://user-images.githubusercontent.com/1094774/77814020-c8327c80-7083-11ea-90c3-26e707c4d55f.png)

That's a 10x improvement 👏 

Closing this. Thank you @visusnet

PR #80, opened 2020-02-21
Performance improvement (single AST traversal run)

visusnet:
This is a continuation of #77.

Please use #79 for discussion.

### DIFF
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -24,7 +24,7 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     const applicableRefactorings: RefactoringWithActionProvider[] = [];
 
     const onCanPeform = (
-      path: t.NodePath<any>,
+      path: t.NodePath,
       refactoring: RefactoringWithActionProvider
     ) => {
       if (isLegacyActionProvider(refactoring.actionProvider)) {
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -58,7 +58,7 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     t.traverseAST(ast, {
       enter: (path: t.NodePath<any>) => {
         refactorings.forEach(refactoring =>
-          this.canPerform(
+          this.visitAndCheckApplicability(
             refactoring,
             path,
             selection,
@@ -82,18 +82,18 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     return applicableRefactorings;
   }
 
-  private canPerform(
+  private visitAndCheckApplicability(
     refactoring: RefactoringWithActionProvider<ActionProvider>,
     path: t.NodePath<any>,
     selection: Selection,
-    onCanPerform: (
+    whenApplicable: (
       matchedPath: t.NodePath<any>,
       refactoring: RefactoringWithActionProvider<ActionProvider>
     ) => void
   ) {
     const visitor: t.Visitor = refactoring.actionProvider.createVisitor(
       selection,
-      path => onCanPerform(path, refactoring),
+      path => whenApplicable(path, refactoring),
       refactoring
     );
 
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -1,9 +1,9 @@
 import * as vscode from "vscode";
 
 import { createSelectionFromVSCode } from "./editor/adapters/vscode-editor";
-import { Selection } from "./editor/selection";
-import { RefactoringWithActionProvider } from "./types";
+import { RefactoringWithActionProvider, isLegacyActionProvider } from "./types";
 import * as t from "./ast";
+import { Selection } from "./editor/selection";
 
 export { RefactoringActionProvider };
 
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -11,21 +16,40 @@ interface Refactoring {
...
+  isPreferred?: boolean;
+  createVisitor: (
+    selection: Selection,
+    onMatch: (path: NodePath<any>) => void,
+    refactoring: RefactoringWithActionProvider
+  ) => Visitor;
+  updateMessage?: (path: NodePath<any>) => void;
+}
+
+interface LegacyActionProvider {
```
nicoespeon: 👍 Nice, so we can progressively adopt

src/types.ts
```diff
@@ -11,21 +16,40 @@ interface Refactoring {
   };
 }
 
+interface ActionProvider {
+  message: string;
+  isPreferred?: boolean;
+  createVisitor: (
+    selection: Selection,
+    onMatch: (path: NodePath<any>) => void,
```
nicoespeon: Doesn't `NodePath` has a default value for the generics?

```suggestion
    onMatch: (path: NodePath) => void,
```

src/types.ts
```diff
@@ -11,21 +16,40 @@ interface Refactoring {
...
+    refactoring: RefactoringWithActionProvider
+  ) => Visitor;
+  updateMessage?: (path: NodePath<any>) => void;
```
nicoespeon: 🍰 Suggestion: make this return a new message string, and update the message with this value from the generic implementation.

Otherwise it's called `updateMessage()` but really it could do anything. I'd prefer to enforce the behavior.

Finally, what do you think about merging this with the `message` one? We could have a `message` attributes that could either be a `string` or a function returning a `string`.

src/refactorings/simplify-ternary/index.ts
```diff
@@ -10,7 +14,12 @@ const config: RefactoringWithActionProvider = {
   },
   actionProvider: {
     message: "Simplify ternary",
-    canPerform: hasTernaryToSimplify
+    createVisitor: createTernaryToSimplifyVisitor,
+    updateMessage(path: t.NodePath<any>): void {
+      this.message = `Simplify ternary (for demo purposes only: ${
```
nicoespeon: Nice 👍 

Don't forget to drop it before we merge, since it's only for demoing it can be done.

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+  private canPerformLegacy(
     refactoring: RefactoringWithActionProvider,
     ast: t.AST,
     selection: Selection
   ) {
     try {
-      return refactoring.actionProvider.canPerform(ast, selection);
+      return (
+        isLegacyActionProvider(refactoring.actionProvider) &&
+        typeof refactoring.actionProvider.canPerform === "function" &&
```
nicoespeon: Is line 102 necessary? Isn't it covered by `isLegacyActionProvider()` already?

```suggestion
        isLegacyActionProvider(refactoring.actionProvider) &&
```

nicoespeon: Alternative suggestion: change the type of `refactoring` to only be a Legacy one + use the guard clause to filter out the non-Legacy ones before calling the function. That will preserve the original behavior of the function and split the responsibilities.

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+  }
+
+  private visit(visitor: any, path: t.NodePath<any>) {
+    const node: t.Node = path.node;
+
+    try {
+      if (typeof visitor[node.type] === "function") {
+        visitor[node.type](path);
+      } else if (typeof visitor[node.type] === "object") {
+        visitor[node.type].enter(path);
```
nicoespeon: 👍 Nice

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+    const visitor: t.Visitor = refactoring.actionProvider.createVisitor(
+      selection,
+      path => onCanPerform(path, refactoring),
+      refactoring
+    );
+
+    this.visit(visitor, path);
+  }
+
+  private visit(visitor: any, path: t.NodePath<any>) {
```
nicoespeon: I think `NodePath` will default on `t.Node` if we don't provide a value to the generics:

```suggestion
  private visit(visitor: any, path: t.NodePath) {
```

nicoespeon: Also: is `visitor` really `any`thing? It seems it should be `t.Visitor`, isn't it?

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     const ast = t.parse(document.getText());
     const selection = createSelectionFromVSCode(range);
 
-    return this.refactorings
-      .filter(refactoring => this.canPerform(refactoring, ast, selection))
-      .map(refactoring => this.buildCodeActionFor(refactoring));
+    const applicableRefactorings: RefactoringWithActionProvider[] = [];
+
+    const onCanPeform = (
+      path: t.NodePath<any>,
```
nicoespeon:
```suggestion
      path: t.NodePath,
```

IMO, it defaults to `t.Node` itself, so we're fine removing it. `any` is actually incorrect.

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     const ast = t.parse(document.getText());
     const selection = createSelectionFromVSCode(range);
 
-    return this.refactorings
-      .filter(refactoring => this.canPerform(refactoring, ast, selection))
-      .map(refactoring => this.buildCodeActionFor(refactoring));
+    const applicableRefactorings: RefactoringWithActionProvider[] = [];
+
+    const onCanPeform = (
```
nicoespeon: I don't like this name much as it doesn't explain what it does, but when it's used. 

Could you inline it directly? I'd prefer not to name it until we find a good name.

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+      path: t.NodePath<any>,
+      refactoring: RefactoringWithActionProvider
+    ) => {
+      if (isLegacyActionProvider(refactoring.actionProvider)) {
```
nicoespeon: As mentioned in the `canPerformLegacy` function: I'd split the legacy and non-legacy refactorings first, and then map them to the appropriate function. That way, we don't have to duplicate the checks everywhere (e.g. here and in `canPerform()`).

What do you think?

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
+      }
+
+      if (refactoring.actionProvider.updateMessage) {
+        refactoring.actionProvider.updateMessage(path);
+      }
+
+      applicableRefactorings.push(refactoring);
+    };
+
+    t.traverseAST(ast, {
```
nicoespeon: 👍 Yes! Here's the single traversal. Nice!

src/action-providers.ts
```diff
@@ -21,18 +21,87 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
...
 
   private canPerform(
+    refactoring: RefactoringWithActionProvider,
+    path: t.NodePath<any>,
+    selection: Selection,
+    onCanPerform: (
+      matchedPath: t.NodePath<any>,
+      refactoring: RefactoringWithActionProvider
+    ) => void
+  ) {
```
nicoespeon: I find it surprising that `canPerform()` is not returning a boolean. I'm not sure I understand how the new implementation is supposed to determine which refactorings can actually be performed.

Isn't this PR breaking the Simplify Ternary quick fix?

src/refactorings/simplify-ternary/index.ts
visusnet: Already removed in one of the other PRs. :)

visusnet:
@nicoespeon I'll check the other suggestions tomorrow (March 3rd, UTC+1 timezone).

nicoespeon:
@visusnet also, I'll be able to give you a hand on this if you need to. Just let me know ;-)

visusnet:
@nicoespeon My team and I are on a team event, i.e. I won‘t be able to do something today. I‘m not sure yet how much there is to do. I‘ll come back to your offer if necessary. Thank you. :)

nicoespeon:
@visusnet no worries. I think I should be able to tackle my comments on Thursday morning (EST time) 😃 

Since I agree with the direction of the PR, my goal is to get this merged while keeping `master` shippable 👍

nicoespeon:
Soooo, I didn't have time to work on this today. I won't have time to do so this week-end and next week neither (conference incoming). That's fine, there's no hurry 👍

src/types.ts
visusnet: I liked your suggestion to merge `updateMessage()` and `message` but they are accessed at different places (`updateMessage()` is called during the AST traversal, `message` is read afterwards). I'll keep the `updateMessage()` function for now but make it return `string`.

src/action-providers.ts
visusnet: Yes, it is a bit strange that `canPerform()` calls a callback. However, moving the callback logic inside leads to `canPerform()` mutating the `refactoring` that is passed which is - to me - more surprising than the function's name. I will think about a different name which makes the function's intent more clear.

visusnet:
@nicoespeon I have updated the PR with most of your suggestions.

src/action-providers.ts
visusnet: It is not really anything but it can mean a lot of things and it has lots of generic stuff. I have tried to use `t.Visitor` instead but this leads to a lot of unreadable code. You can give it a try. Your typescript magic might be stronger than mine. 😄

src/refactorings/simplify-ternary/index.ts
```diff
@@ -10,7 +14,10 @@ const config: RefactoringWithActionProvider = {
   },
   actionProvider: {
     message: "Simplify ternary",
-    canPerform: hasTernaryToSimplify
+    createVisitor: createTernaryToSimplifyVisitor,
+    updateMessage(path: t.NodePath): string {
+      return `Simplify ternary (for demo purposes only: ${path.node.type})`;
```
nicoespeon: image

I can see it's working 👍 

I'll update the message so it's OK to deploy once on master. Good job 👍

PR #81, opened 2020-02-21
Refactored most refactorings to use the createVisitor pattern.

visusnet:
This is a continuation of #77 and #79.

PR #82, opened 2020-02-21
Performance improvement (use visitor)

visusnet:
#77 + #80 + #81 + (most) refactorings (where possible) use the visitor pattern.

See #79.


nicoespeon:
Everything looks good. I removed a few type casts because it defeats TS hints. Usually, it's just a matter of proper types that aren't passed, which I fixed.

When passing a proper type was not super convenient (e.g. it would involve creating a custom type), I just considered that could eventually happen and duplicated the check to make the code type safe. So yep, we might do a non-necessary check at runtime, but I prefer that as a compromise rather than loosing the type safety 😃 

I'll check the perf changes and document that, then I'll merge the PR.

Thank you a lot @visusnet 🙏

nicoespeon:
@visusnet I did some perf analysis on a big file and I found interesting stuff.

First of all, I realized that the `reactConvertToPureComponent` Quick Fix takes a looooong time. This is because we execute the whole refactoring on the code and compare the output. Plus, we don't control the code that's being executed, and it's clearly not optimized for such usage: https://github.com/nicoespeon/abracadabra/blob/5c4d2d56fc048f9b660d1508968a34bc98c0ff1b/src/refactorings/react/convert-to-pure-component/convert-to-pure-component.ts#L44-L47

The time taken is considerable:

image

1.70s out of 2.06s, that's 83% of the time spent!

Thus, we'll remove this Quick Fix. People will still be able to run this refactoring through the command palette.

---

Now, taking this refactoring out of the Quick Fixes, here's what we have before:

image

It took 1.48s to execute.

What we have after:

image

It took 1.02s to execute. That's a very cool improvement, 33% faster 👏 

There's one refactoring left on the old pattern: `hasDeadCode`. **I'll try to port this one too**.

---

I also fixed a few issues with "Merge if statements" and "Add braces to if statement" quick fixes. I put that in the PR to avoid merge conflicts. I updated the changelog.

For some reason, I don't know why CI is not running on my commits 🤷‍♂ 

Finally, **we have an issue to fix before we can merge this**. Because we don't stop the visitor, we can have multiple matches for the same refactoring:

image

_On a side-note, I'm considering dropping this refactoring in particular since it's now part of VS Code and the DX of the native one is great!_

## Todos

- [ ] Drop React convert to pure component Quick Fix
- [ ] Port `hasDeadCode` to the new pattern
  - Maybe drop the old pattern if it's not used anymore? Can be done in a follow-up PR.
- [ ] Fix duplicate Quick Fixes

nicoespeon:
There we go, no more "Convert to Pure Component" quick fix, nor legacy refactoring one:

image

Even though there might be volatility between each perf measure, the flame graph clearly shows that it's faster now. We're only traversing the AST once, that makes a great difference 👍 

I can feel the before/after on that large JS file (it's a 2.3k LOC file, which is not the largest, but definitely too big already).

nicoespeon:
@visusnet everything is good. I was able to tackle all the changes I mentioned. We're now completely using the new way, so I dropped the "Legacy" stuff 🎉 

If we consider the "Convert to Pure Component" impact, the extension is 10x faster to run on large file on my computer. If we omit this one to focus on your specific work, it goes 2x faster, which is **a great achievement** 🚀 

Also, it removed duplication across many refactorings, making them simpler 👍 

Big thanks for that. I'm gonna merge this and close the related issue.

Issue #87, opened 2020-03-11
ActionProvider throws if code can't be parsed


tjx666:
## Describe the bug

When I write following code:

![screenshot_2020-03-11_18-44-56](https://user-images.githubusercontent.com/41773861/76408769-707bec00-63c8-11ea-9377-dc9d59b44fbf.png)

error message: 
```
Unexpected token (29:4): SyntaxError: Unexpected token (29:4)
	at Object.raise (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:657452)
	at Object.unexpected (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:677107)
	at Object.parseExprAtom (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:698149)
	at Object.parseExprAtom (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:575944)
	at Object.parseExprSubscripts (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:690154)
	at Object.parseMaybeUnary (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:689746)
	at Object.parseMaybeUnary (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:646315)
	at Object.parseExprOps (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:687156)
	at Object.parseMaybeConditional (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:686688)
	at Object.parseMaybeAssign (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:685576)
	at Object.parseMaybeAssign (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:645779)
	at Object.parseObjectProperty (c:\Users\YuTengjing\.vscode\extensions\nicoespeon.abracadabra-3.2.1\out\extension.js:1:706149)
[...]

PR #90, opened 2020-03-12
Upgrade @babel/parser

### LINKED ISSUE #75: Update @babel/parser
Deno has top level await enabled, and so does TypeScript 3.8.

At the moment any code with this in stops abracadabra working. I think it could be cured by updating https://www.npmjs.com/package/@babel/parser to the latest, it seems it supports top level await now, but I don't know if you need to enable it? I have not used Babel. Cheers!

nicoespeon:
Fixes #75
Maybe related to #87 

- Upgrade babel to latest version (7.8.7)
- Upgrade jest to latest version (25.1.0)
- Fix type issues and failing tests because of breaking changes
- Handle top-level awaits
- Add all compatible and missing babel plugins when parsing code to handle a maximum of scenario

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "10.0.0"
+  - "10.10.0"
 sudo: false
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.5.1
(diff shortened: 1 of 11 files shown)


nicoespeon:
I don't understand why CI is failing. It might have to do with packages and cache. Working on it 👀

nicoespeon:
There you go 😄

Issue #93, opened 2020-04-07
New Refactoring: Convert Switch to If/Else


delaaxe:
This would help me and others a lot. Would it be complex to implement, assuming we already have if/else to switch?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

nicoespeon:
Sure, that will be a good idea indeed. So it's easy to go back and forth between the 2 patterns 👍 

We need to [create a new refactoring](https://github.com/nicoespeon/abracadabra/blob/master/CONTRIBUTING.md#create-a-new-refactoring) named `convertSwitchToIfElse`.

Then implement it, using tests to guide the progression. I'm not sure how complex it would be, but tests are really helpful to implement refactorings step by step. I'm sure we can get a basic scenario working pretty quickly 👍 

I'd put this one as a "Good First Contribution" if anyone wants to take a stab at it. Read [the CONTRIBUTING guide](https://github.com/nicoespeon/abracadabra/blob/master/CONTRIBUTING.md#create-a-new-refactoring), ask questions and open a draft PR if you need help!

nicoespeon:
Released in [v4.2.0](https://github.com/nicoespeon/abracadabra/releases/tag/4.2.0) 🎉

PR #102, opened 2020-04-25
New Refactoring: Convert Switch to If/Else

delaaxe:
Implements #93

### DIFF
--- a/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
+++ b/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
@@ -37,6 +37,43 @@ default:
 }`
       },
       {
+        description: "empty fall-throughs",
+        code: `switch (name) {
+case "Jane":
+  sayHelloToJane();
+  break;
+case "Joe":
+case "Joseph":
+  sayHelloToJoe();
+  break;
+case "John":
+case "Johnny":
+case "Johnie":
+  sayHelloToJohn();
+  break;
+case "Jack":
+case "Jacob":
+case "Jackie":
+case "Jake":
+  sayHelloToJack();
+  break;
+default:
+  sayHello();
+  break;
+}`,
+        expected: `if (name === "Jane") {
+  sayHelloToJane();
+} else if (name === "Joe" || name === "Joseph") {
+  sayHelloToJoe();
+} else if (name === "John" || (name === "Johnny" || name === "Johnie")) {
+  sayHelloToJohn();
+} else if (name === "Jack" || (name === "Jacob" || (name === "Jackie" || name === "Jake"))) {
+  sayHelloToJack();
+} else {
+  sayHello();
+}`
+      },
+      {
         description: "convert the selected switch only",
         code: `switch (name) {
 case "Jane":
@@ -47,10 +84,13 @@ default:
   break;
 }
 
-if (name === "John") {
+switch (name) {
+case "John":
   sayHelloToJohn();
-} else {
+  break;
+default:
   sayHello();
+  break;
 }`,
         expected: `if (name === "Jane") {
   sayHelloToJane();
@@ -58,10 +98,13 @@ if (name === "John") {
   sayHello();
 }
 
-if (name === "John") {
+switch (name) {
+case "John":
   sayHelloToJohn();
-} else {
+  break;
+default:
   sayHello();
+  break;
 }`
       },
       {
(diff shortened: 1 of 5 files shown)


src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
```diff
@@ -0,0 +1,239 @@
...
+        expected: `if (name === "Jane") {
+  sayHelloToJane();
+} else if (name === "John") {
+  sayHelloToJohn();
+} else {
+  sayHello();
+}`
+      },
+      {
+        description: "empty fall-throughs",
```
nicoespeon, line 40: 👍 For thinking about this scenario

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      expect(result).toBe(expected);
+    }
+  );
+
+  testEach<{ code: Code; selection?: Selection }>(
+    "should not convert",
+    [
+      {
+        description: "non-empty case without break",
```
nicoespeon, line 186: 👍 Great!

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
```diff
@@ -0,0 +1,239 @@
...
+  async function doConvertSwitchToIfElse(
+    code: Code,
+    selection: Selection
+  ): Promise<Code> {
+    const editor = new InMemoryEditor(code);
+    editor.showError = showErrorMessage;
+    await convertSwitchToIfElse(code, selection, editor);
+    return editor.code;
+  }
+});
```
nicoespeon, line 239: 👏 Solid tests, that's really helpful. Thank you!

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
```diff
@@ -0,0 +1,172 @@
...
+  });
+
+  return result;
+}
+
+function convert(node: t.SwitchStatement): t.SwitchStatement | t.IfStatement {
+  try {
+    return convertNode(node);
+  } catch (err) {
+    console.log(err);
```
nicoespeon, line 81: We don't want to `console.log()` anything generally. If there's an error, it should be explicit to the user.

But we don't want such error to be explicit. Nor we want to log it: this refactoring will be checked anytime the cursor is moved (so we can propose the quick fix to the user if it can be performed). So that would log an error each time we move the cursor in a switch statement that can't be converted.

I think we mostly use the error to control the flow here (that is, to return an `IfStatement` instead of the `SwitchStatement`). 

I think we could use the NullObject pattern instead. It will probably simplify this.

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
```diff
@@ -0,0 +1,172 @@
...
+    );
+  }
+
+  if (statements.length === 1) {
+    return firstStatement;
+  }
+
+  for (let i = 0; i < statements.length - 1; i++) {
+    const statement = statements[i];
+    if (statement.type === "IfStatement") {
```
nicoespeon, line 145: I think we should be able to use `t.isIfStatement(statement)` instead.

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
delaaxe: Ahah I commit the console.log by mistake it was only there for debugging purposes... Good catch

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
delaaxe: Didn't know about that, thanks

nicoespeon:
@delaaxe I didn't published a release yet but will do today. You'll be able to see the refactoring when this is done 👍 

If you're curious, here's the commit: a1087b5c436f7f0e9b3bd6d1f698d75728b830bc

delaaxe:
I meant that while I was testing with my local build. I saw you add some confit later on, maybe this should be added to the code generator?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #103, opened 2020-04-30
Switch to If/Else: continued

delaaxe:
This PR outputs more natural if statements when all switch cases end with return statements.

### DIFF
--- a/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
+++ b/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
@@ -165,11 +165,34 @@ default:
 }`,
         expected: `if (name === "Jane") {
   return sayHelloToJane();
-} else if (name === "John") {
+}
+
+if (name === "John") {
   return sayHelloToJohn();
-} else {
+}
+
+return sayHello();`
+      },
+      {
+        description: "with return statements and fallthrough",
+        code: `switch (name) {
+case "Jane":
+  return sayHelloToJane();
+case "John":
+case "Johnny":
+  return sayHelloToJohn();
+default:
   return sayHello();
-}`
+}`,
+        expected: `if (name === "Jane") {
+  return sayHelloToJane();
+}
+
+if (name === "John" || name === "Johnny") {
+  return sayHelloToJohn();
+}
+
+return sayHello();`
       }
     ],
     async ({ code, selection = Selection.cursorAt(0, 0), expected }) => {
--- a/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
+++ b/src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
@@ -145,8 +145,7 @@ function linkIfStatements(node: t.SwitchStatement, statements: t.Statement[]) {
 
   const allEndWithReturn = node.cases.every(
     caseNode =>
-      t.isReturnStatement(last(caseNode.consequent)) ||
-      caseNode.consequent.length === 0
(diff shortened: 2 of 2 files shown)


src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
```diff
@@ -24,7 +24,11 @@ function updateCode(ast: t.AST, selection: Selection): t.Transformed {
   return t.transformAST(
     ast,
     createVisitor(selection, (path, convertedNode) => {
-      path.replaceWith(convertedNode);
+      if (t.isBlockStatement(convertedNode)) {
```
delaaxe, line 27: Returning a block node to convey that there are multiple nodes to replace may not be the cleanest solution but it worked

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
```diff
@@ -34,7 +38,7 @@ function createVisitor(
   selection: Selection,
   onMatch: (
     path: t.NodePath<t.SwitchStatement>,
-    convertedNode: t.IfStatement | t.SwitchStatement
+    convertedNode: t.Statement
```
delaaxe, line 41: This could also be
```suggestion
    convertedNode: t.IfStatement | t.SwitchStatement | t.Statement[]
```
but maybe no

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
```diff
@@ -165,11 +165,34 @@ default:
...
   return sayHelloToJane();
-} else if (name === "John") {
+}
+
+if (name === "John") {
   return sayHelloToJohn();
-} else {
+}
+
+return sayHello();`
```
nicoespeon, line 174: 👍 Nice, that's a better output for return statements indeed!

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
```diff
@@ -165,11 +165,34 @@ default:
...
+}`,
+        expected: `if (name === "Jane") {
+  return sayHelloToJane();
+}
+
+if (name === "John" || name === "Johnny") {
+  return sayHelloToJohn();
+}
+
+return sayHello();`
```
nicoespeon, line 195: 👍 Tests look great!

What if you have a mix of return statements and non-return?

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
nicoespeon: That's fine, for the moment it works. We can still refactor the code later if we need to change it, thanks to the tests you've written it's easy to do so 👍

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.ts
nicoespeon: It's OK to return `t.Statement` here I think.

src/refactorings/convert-switch-to-if-else/convert-switch-to-if-else.test.ts
nicoespeon: I see there's `allEndWithReturn` in the code so I guess it will behave as expected.

Can you still add a test to illustrate?

nicoespeon: I'll merge and do it, so I can release a new version today ;)

Thanks for your hard work 🙏

PR #108, opened 2020-05-18
[Extract Generic Type] Handle function declaration

nicoespeon:
Duplicate interface implementation to handle function declarations too.

Will refactor the duplication after the release. I need this refactoring to work on functions, and there's no reason not too 😉


src/refactorings/extract/extract-generic-type/extract-generic-type.ts
```diff
@@ -200,3 +216,96 @@ class SelectedInterfaceOccurrence extends InterfaceOccurrence {
     );
   }
 }
+
+class FunctionOccurrence implements Occurrence {
```
nicoespeon, line 220: Most of this code is duplicated from InterfaceOccurrence. I'll refactor that.

Issue #107, opened 2020-05-17
Convert a `let` to `const` for a variable declaration where the context allows it


nickebbitt:
## Is this request related to a problem? Please describe.

Favouring immutability, it would be good to have a quick fix to convert a variable declaration from `let` to `const` if the scope and usage of the variable allows it.

## Describe the solution you'd like

Given the following code:
```
function getValue(): string {
   let someVariable = 'value';
   return someVariable;
}
```
A quick-fix on `someVariable` would allow conversion to this:
```
function getValue(): string {
   const someVariable = 'value';
   return someVariable;
}
```
Whereas it wouldn't be available for:
```
function getValue(): string {
   let someVariable = 'value';
   someVariable = 'anotherValue';
   return someVariable;
}
```

nicoespeon:
This would be a nice feature. Obviously, the tricky part of it is to detect usages of the identifier in scope to detect whether or not it's re-assigned.

The transformation part is quite easy to implement.

I'll mark it as a good first issue. I can provide guidance on a draft PR if one needs.

nickebbitt:
Should we also look to handle `var` for this feature? Seems like it makes sense.

nicoespeon:
`var` is trickier because of the scope, but yeah it would make sense!

I wouldn't mind starting with let / const and iterate to handle var later on.

chemitaxis:
I can do it, can you please explain a little about how to solve? Thanks!

nickebbitt:
Hey @chemitaxis, sorry if it wasn't obvious but I've started work on this. I wouldn't want this to stop you though contributing though.

Maybe @nicoespeon has another issue in mind that would be a good one to get started on?

nickebbitt:
@chemitaxis a great place to start anyway is to work through the [contributing guide](https://github.com/nicoespeon/abracadabra/blob/master/CONTRIBUTING.md).

nicoespeon:
Shipped in a brand new release ✨

PR #109, opened 2020-05-26
Convert a `let` to `const` for a variable declaration

nickebbitt:
Implements #107 

Doesn't currently convert where there are multiple variable declarators on the same line e.g. 

```
let someVariable, otherVariable = 'value';
someVariable = 'anotherValue';
```

This would require splitting the declarations which is doable but I felt maybe an iteration once we have a working baseline.


nickebbitt:
@nicoespeon interested in your feedback on the impl and can you think of any edge cases I've not covered that we'd like tests for?

src/refactorings/convert-let-to-const/convert-let-to-const.ts
```diff
@@ -0,0 +1,88 @@
+import { Editor, Code, ErrorReason } from "../../editor/editor";
+import { Selection } from "../../editor/selection";
+import * as t from "../../ast";
+import { isLet } from "../../ast";
+import { Binding } from "@babel/traverse";
```
nickebbitt, line 5: @nicoespeon I'm currently importing `@babel/traverse` directly here however was wondering if maybe we'd prefer this behaviour to live in the `ast` module?

src/refactorings/convert-let-to-const/convert-let-to-const.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  let otherVariable = 'otherValue';
+}`,
+        expected: `{
+  const someVariable = 'someValue';
+  let otherVariable = 'otherValue';
+}`
+      },
+      {
+        description:
+          "multiple variables delcared seperately, other one mutated",
```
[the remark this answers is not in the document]
nickebbitt: yes that makes sense, my bad, I'd interpreted the selection to have a wider scope (i.e. all variables on that line) than it does, rather than actually a single selected variable.

I agree with your thoughts so will make the changes 👍

src/refactorings/convert-let-to-const/convert-let-to-const.test.ts
```diff
@@ -0,0 +1,160 @@
...
+        description: "non-mutated variable declared in a block",
+        code: `{
+  let someVariable = 'value';
+}`,
+        expected: `{
+  const someVariable = 'value';
+}`
+      },
+      {
+        description: "multiple non-mutated variables declared together",
```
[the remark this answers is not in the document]
nickebbitt: Thinking about this one, I kind of overlooked the fact that the refactor would leave us with invalid code due to the `'const' declarations must be initialized.` error.

At least initially I'm going to prevent the refactor for this case.

Eventually, I think the solution for this is to move the selected variable onto its own line leaving the other variable unaffected by the refactor.

@nicoespeon does that feel like a reasonable decision?

src/refactorings/convert-let-to-const/convert-let-to-const.ts
nickebbitt: @nicoespeon I think this is the only change left to make, I looked at the `ast` module but unsure where this should go.

Does it fit in one of the existing files or require a new one?

nicoespeon: It has to do with `@babel/traverse` so I'd expose it here: https://github.com/nicoespeon/abracadabra/blob/3326ba97cfe1a8049f741148fc12c19ed308729a/src/ast/transformation.ts#L2

src/refactorings/convert-let-to-const/convert-let-to-const.test.ts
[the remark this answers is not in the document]
nicoespeon: Yep, I think it's the best decision. We shouldn't turn an un-initialized `let` into `const` or that would be invalid. Of course it makes the multi-variables trickier to deal with, but it's the right thing to do 👍

src/refactorings/convert-let-to-const/convert-let-to-const.ts
nickebbitt: Done 👍 

Unless you've any other feedback this hopefully we're good to merge :)

nickebbitt:
@nicoespeon I added some details to the README, not sure if the section is exactly right? https://github.com/nicoespeon/abracadabra/pull/109/commits/ea5c9592b73ffe28111b69740bd3f482f1902537

nicoespeon:
@nickebbitt I think the section is just right 👌 

I'll have a final look tomorrow but that looks great to me already 👏

src/refactorings/convert-let-to-const/convert-let-to-const.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        code: `{
+  let someVariable = 'value';
+}`,
+        selection: Selection.cursorAt(1, 7),
+        expected: `{
+  const someVariable = 'value';
+}`
+      },
+      {
+        description: "only the selected non-mutated variable",
```
nicoespeon, line 35: 👏

src/refactorings/convert-let-to-const/convert-let-to-const.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function createVisitor(
+  selection: Selection,
+  onMatch: (path: t.NodePath<t.VariableDeclarator>) => void
+): t.Visitor {
+  return {
+    VariableDeclarator(path) {
+      const { node } = path;
+      if (!selection.isInsideNode(node)) {
```
nicoespeon, line 47: 👍

src/refactorings/convert-let-to-const/convert-let-to-const.ts
```diff
@@ -0,0 +1,71 @@
...
+  };
+}
+
+function isSingleLetVariableDeclaration(
+  node: t.Node
+): node is t.VariableDeclaration {
+  return t.isLet(node) && node.kind === "let" && node.declarations.length === 1;
+}
+
+function variableCanBeConst(
```
nicoespeon, line 61: 👍

PR #111, opened 2020-06-18
Extract variable/template literal partial selection

### LINKED ISSUE #98: Extract Variable from a subset of string or template literal
## Describe what's missing

It would be nice to be able to extract selected text subset of a string or template literal.

## How would it work

For example, with this code:

```js
const message = "[*] Mailing list worker started (attached to web app)";
```

If we select `Mailing list worker` and run Extract Variable, the code should be:

```js
const extracted = "Mailing list worker";
const message = `[*] ${extracted} started (attached to web app)`;
```

Same for template literals.

## Additional information

We should probably handle a bunch of edge cases:
- if we select the whole string, we should extract the whole string and not have \`${extracted}\`
- if we don't select something and we just have the cursor inside the string, it should still extract the whole string
- if selection is not just a string (e.g. in a template literal, we could select the interpolated variables), then it should extract the whole string too

nicoespeon:
Fixes #98 

It works on strings and template literals. 

If you select explicitly a part of the string, it will extract just that. 
It will still extract the whole string if you just put your cursor inside, or if you select the whole string, or if your selection can't be extracted.

It doesn't work with multi-lines template literals yet. But we thought it was good enough to be released since it covers the most frequent use-cases!

extract-substring


src/refactorings/extract/extract-variable/parts.ts
```diff
@@ -0,0 +1,33 @@
+import { Code } from "../../../editor/editor";
+import { Selection } from "../../../editor/selection";
+import { Position } from "../../../editor/position";
+
+export { Parts };
+
+class Parts {
```
nicoespeon, line 7: This is the code that should be enriched if we want to handle multi-lines selections.

For now, it gracefully falls back on extracting the whole string instead.

src/refactorings/extract/extract-variable/parts.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const { before, selected, after } = new Parts(code, selection, offset);
+
+    expect(before).toBe("Hello ");
+    expect(selected).toBe("world");
+    expect(after).toBe("!");
+  });
+
+  // This doesn't work yet for multi-lines code because we don't support it.
+  it.skip("should work on a single-line selection of a multi-lines code", () => {
```
nicoespeon, line 31: This is the test that should be implemented if we want to make multi-lines selection works.

Issue #113, opened 2020-06-26
"Convert function declaration to expression or arrow function" (and vice versa)


OliverJAsh:
Hey @nicoespeon. I just watched [your fantastic workshop](https://www.youtube.com/watch?v=udNV7zPN1H8), thank you so much.

In the workshop you demonstrate a refactoring for converting function declarations to arrow functions. Do you think we could add something like this to Abracadabra?

We would probably want to support conversions both ways, between declarations and expressions. When converting to an expression, we should support both non-arrow and arrow formats.

`vscode-javascript-booster` has something along these lines:

- https://github.com/xsburg/vscode-javascript-booster/blob/c2f1d10a2c256e8f4765ab7e6dc88e8a6629fc26/server/src/codemods/convert-arrow-fn-to-regular-fn.ts
- https://github.com/xsburg/vscode-javascript-booster/blob/c2f1d10a2c256e8f4765ab7e6dc88e8a6629fc26/server/src/codemods/convert-fn-decl-to-arrow-fn.ts
- https://github.com/xsburg/vscode-javascript-booster/blob/c2f1d10a2c256e8f4765ab7e6dc88e8a6629fc26/server/src/codemods/convert-fn-expr-to-arrow-fn.ts

Although when I tried it [they seemed to lose TypeScript generics](https://github.com/xsburg/vscode-javascript-booster/issues/33). I'm sure Abracadabra could do a better job 😉

nicoespeon:
Hey @OliverJAsh 👋 

Thanks for the kind words! And yes, that's funny I presented an example that's not even in the extension 😉 

That's totally do-able with Abracadabra. It's not hard to preserve types, we already have some TS-specific refactorings so there's tooling to deal with types.

The only gotcha I can think of is hoisting. This is valid:

```js
doSomething();

function doSomething() {}
```

This is not, because the function declaration isn't hoisted anymore, so it's used before it's declared:

```js
doSomething();

const doSomething = () => {};
```

The refactoring should check that there's no usage of the function above the function declaration, at least. If refactoring can't be made safely, it shouldn't be proposed (this is the way it works today, although in the future we could let the user know and deal with it instead).

For the record, Webstorm warns the user when the refactoring would be invalid:

![image](https://user-images.githubusercontent.com/1094774/85965406-9aaee500-b98a-11ea-81f9-0b0891f2e504.png)

nicoespeon:
In case it's not clear: I think this would be a nice refactoring. 

[...]

PR #114, opened 2020-06-29
Convert function declaration to arrow function

OliverJAsh:
Re. https://github.com/nicoespeon/abracadabra/issues/113

@nicoespeon ~The tests pass but the code action does not appear. What am I missing?~ Figured it out, the import in `extension.ts` was missing. Perhaps this should be part of the code generation?

P.S. your contributing guide is fantastic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "onCommand:abracadabra.addBracesToIfStatement",
     "onCommand:abracadabra.bubbleUpIfStatement",
     "onCommand:abracadabra.convertForToForeach",
+    "onCommand:abracadabra.convertFunctionDeclarationToArrowFunction",
     "onCommand:abracadabra.convertIfElseToTernary",
     "onCommand:abracadabra.convertIfElseToSwitch",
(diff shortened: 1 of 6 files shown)


nicoespeon:
> Figured it out, the import in `extension.ts` was missing. Perhaps this should be part of the code generation?

Good point, that's something missing.

Actually, I didn't put it in the code generation because I order the imports in `extension.ts` alphabetically. So I do that by hand. 

There's also the `package.json` that should be configured. For this, at some point, I think I'll make it generate automatically so there's less manual work and potential errors. Maybe we could even make the `extension.ts` smarter so we don't really have to touch it again and it's all about configuration in each individual refactoring 🌈 ☀️ 

I think the simplest way to solve this today is to mention it in the contributing guide, to explain the need for import + order alphabetically. That will at least help future contributors waste less time.

OliverJAsh:
> I think the simplest way to solve this today is to mention it in the contributing guide, to explain the need for import + order alphabetically. That will at least help future contributors waste less time.

Done in #116

PR #115, opened 2020-06-29
Refactor idea and proof of concept: "Wrap (expression) in function call"

OliverJAsh:
I find myself constantly needing to wrap an expression in a function call, so it would be useful to have a refactor for this. This is how I was doing it before, manually:

Screen Recording 2020-06-29 at 17 49 21

And this is with the new refactoring command:

Screen Recording 2020-06-29 at 17 49 49

Note I couldn't get the code action to appear. It seemed to be because of this line: https://github.com/nicoespeon/abracadabra/blob/878598a530680c3706037170973c9fc416f15b0a/src/action-providers.ts#L100

`visitorNode` was resolving to `undefined`. This is because I'm traversing the AST for `Expression`, but no nodes have this type directly—it's just a form of grouping, IIUC.

I'm aware that this refactor creates temporarily broken code, because there will most likely be no function named `wrapped`. This is why I've made the refactor immediately select the callee, so it can be replaced, with the help of VS Code auto completion.

I've added some more use cases to the tests.

WDYT? Maybe this isn't something that should live in Abracadabra, but your feedback would be appreciated very much. I've learnt so much from playing with this extension! ❤️

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "jscodeshift": "0.6.4",
     "pluralize": "8.0.0",
     "react-codemod": "5.1.1",
-    "recast": "0.18.1"
+    "recast": "0.19.1"
   },
   "activationEvents": [
     "onCommand:abracadabra.addBracesToArrowFunction",
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "onCommand:abracadabra.simplifyTernary",
     "onCommand:abracadabra.splitDeclarationAndInitialization",
     "onCommand:abracadabra.splitIfStatement",
+    "onCommand:abracadabra.wrapInFunctionCall",
     "onLanguage:javascript",
     "onLanguage:javascriptreact",
     "onLanguage:typescript",
@@ -279,6 +280,11 @@
       {
         "command": "abracadabra.splitIfStatement",
         "title": "Split If Statement",
+        "category": "Abracadabra"
+      },
+      {
+        "command": "abracadabra.wrapInFunctionCall",
+        "title": "Wrap In Function Call",
         "category": "Abracadabra"
       }
     ],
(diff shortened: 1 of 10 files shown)


OliverJAsh:
Another example (using `pipe` from `fp-ts`):

Screen Recording 2020-06-29 at 17 51 02

nicoespeon:
I **love** this proposition ❤️ 

That would be so convenient! Thanks for pushing this.

Of course, as you mentioned, I don't think this should be in Abracadabra. The reason is that Abracadabra focuses on a clear definition of refactorings, so we don't start adding too many things inside the extension. Also, it's easy to tell if something is a bug or not (= if it changes code behavior).

However, I have good news. There was a couple of code transformations I wanted to perform but didn't fit Abracadabra scope. So I created a different extension which focuses on creating things that don't exist yet (= generate code for you, quickly). This is [Hocus Pocus](https://github.com/nicoespeon/hocus-pocus/). 

The codebase is younger and has a bit less tooling than Abracadabra, but the approach is similar and there's probably enough. I'd suggest to port your POC to this extension instead because it perfectly fits the scope 👌 

Would you mind replicate your change here? Let me know if you have any blocker with the tooling being different (maybe things would be missing).

OliverJAsh:
@nicoespeon Thanks for the feedback! I just had a quick go at replicating the change in Hocus Pocus but I need to dig deeper, so I'll come back to this shortly.

nicoespeon:
Next steps:

- let's have a go at migrating this to Hocus Pocus and see what are the blockers, so we tackle them together

nicoespeon:
Would be better fit in https://github.com/nicoespeon/hocus-pocus/

Closing this PR in the meantime, as discussed together

PR #118, opened 2020-06-30
Upgrade Babel parser to 7.9.4

OliverJAsh:
This seemed to fix the bug I mentioned over Twitter DMs, whereby `loc` was `null` for the call expression in `export default a(1);`.

I chose 7.9.4 because:

- that's the version used on AST explorer and I wasn't able to reproduce the bug there
- I had type errors when I tried to upgrade to the latest versions (7.10.x)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -260,15 +260,10 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
-"@babel/parser@7.9.4":
+"@babel/parser@7.9.4", "@babel/parser@^7.1.0", "@babel/parser@^7.1.6", "@babel/parser@^7.7.5", "@babel/parser@^7.8.6", "@babel/parser@^7.8.7", "@babel/parser@^7.9.0":
   version "7.9.4"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -227,6 +227,11 @@
   dependencies:
     "@babel/types" "^7.8.3"
 
+"@babel/helper-validator-identifier@^7.10.4", "@babel/helper-validator-identifier@^7.9.0":
+  version "7.10.4"
```
OliverJAsh, line 231: Ahh I have no idea why we're getting 7.10 here. Yarn deduping is the worst.

nicoespeon:
I hoped that was a bug with the parser and that an upgrade would fix it. Thanks for doing that 👍 

I can merge confidently since CI is all green 👌

nicoespeon:
@all-contributors add OliverJAsh for ideas,bug,code,doc

(I'm sure it won't work, I'll fix this so it works again)

yarn.lock
nicoespeon: No idea! As long as tests are green and bug is fixed, that's legit to me.

nicoespeon:
@all-contributors add OliverJAsh for ideas,bug,code,doc

nicoespeon:
@all-contributors please add OliverJAsh for ideas, bug, code and doc

PR #181, opened 2020-10-20
fix: "Extract Class" refactoring from class that implements multiple interfaces

justerest:
I found this bug in angular components and just fixed it.

### DIFF
--- a/src/refactorings/extract-class/TypescriptClassNode.test.ts
+++ b/src/refactorings/extract-class/TypescriptClassNode.test.ts
@@ -1,0 +1,16 @@
+import { TypescriptClassNode } from "./TypescriptClassNode";
+
+describe(TypescriptClassNode.name, () => {
+  describe(
+    TypescriptClassNode.prototype.createClassNodeWithSameInstanceMembers.name,
+    () => {
+      it("should clone class node without interfaces", () => {
+        const source = "class Source implements ISource, ISource2 {}";
+        const expected = "class Clone {}";
+        const node = TypescriptClassNode.from(source);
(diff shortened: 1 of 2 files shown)


src/refactorings/extract-class/TypescriptClassNode.test.ts
```diff
@@ -0,0 +1,16 @@
+import { TypescriptClassNode } from "./TypescriptClassNode";
+
+describe(TypescriptClassNode.name, () => {
+  describe(
+    TypescriptClassNode.prototype.createClassNodeWithSameInstanceMembers.name,
+    () => {
+      it("should clone class node without interfaces", () => {
+        const source = "class Source implements ISource, ISource2 {}";
```
nicoespeon, line 8: 👍 Nice!

src/refactorings/extract-class/TypescriptClassNode.ts
```diff
@@ -59,12 +59,16 @@ export class TypescriptClassNode implements ClassNode {
...
     clone.node.getStaticMembers().forEach((field) => field.remove());
     clone.node.getDecorators().forEach((field) => field.remove());
-    clone.node
-      .getImplements()
-      .forEach((_, i) => clone.node.removeImplements(i));
+    clone.removeAllImplements();
     return clone;
   }
 
+  private removeAllImplements(): void {
```
nicoespeon, line 66: 👍 Thanks for encapsulating these details in a method. Name is perfect.

nicoespeon (approved):
PR is great! Thanks for finding and fixing this edge case 👏

PR #244, opened 2021-02-05
Split declaration initialization / Fix TS Annotation Duplication

### LINKED ISSUE #232: Split declaration and initialization duplicates type in ts declarations
## Describe the bug

For variable declarations in Typescript, type annotations are duplicated by the "split declaration and initialization" refactoring.

## How to reproduce

1. Initial code:
```ts
let index: number = 0;
```

2.  Apply "split declaration and initialization" refactoring.

3. Output:
```ts
let index: number;
index: number = 0;
```

## Expected behavior

Should be only one type annotation:
```ts
let index: number;
index = 0;
```

## Investigation & Solution
The AST hierarchy for type annotations looks like this:
```js
{
  type: 'VariableDeclaration',
  declarations: [
    {
      type: 'VariableDeclarator',
      id: {
        type: 'Identifier',
        typeAnnotation: { // <= Shouldn't be copied
          type: 'TSTypeAnnotation',
          typeAnnotation: 'TSNumberKeyword'
        }
      }
    }
  ]
}
```

The `typeAnnotation` field on the `Identifier` node is important.

When an `ExpressionStatement` node is created with the `Identifier` copied from the declaration, the `typeAnnotation` field on `Identifier` should not be copied too.

Note that for `Identifiers` without annotation, the `typeAnnotation` field is not set to `undefined`, but is undeclared. So, the `typeAnnotation` field shouldn't be set to `undefined` when copied, but deleted instead.

A possible solution is to delete the `typeAnnotation` field after the Identifier node is copied.

## Additional information

- Version of the extension impacted: 4.13.1

- Note that the AST explorer doesn't support Typescript, so to look at the ASL for Typescript code I made a set up like this:
```js
import parser from "@babel/parser";

const code = `let index: number = 0;`;

const ast = parser.parse(code, {
  plugins: ["typescript"]
});

console.log(ast.program.body[0].declarations[0].id)
```

iulspop:
Closes #232 

Adds a test to cover split declaration with TS annotation and fixes the duplication issue.

split-declaration-initialization-ts

I'm not sure if my implementation is readable. I felt the need to add a comment to clarify why I create a new Identifier.
It's necessary, because if I reassign the `typeAnnotation` property to null, it mutates the Identifier node in the `variableDeclaration` too. I want the annotation to remain in the declaration, so I need to create a new Identifier and modify that.

Feel free to improve the implementation!

### DIFF
--- a/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.test.ts
+++ b/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.test.ts
@@ -62,6 +62,13 @@ firstName = "Jane";
 lastName = "Doe";`
       },
       {
+        description: "declarations with type annotations",
+        code: `const firstName: string = "Jane", age: number = 90;`,
+        expected: `let firstName: string, age: number;
+firstName = "Jane";
+age = 90;`
+      },
+      {
         description: "nested declaration, cursor on wrapper",
         code: `const getLastName = () => {
   const lastName = "Doe";
--- a/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
+++ b/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
@@ -33,8 +33,8 @@ function updateCode(ast: t.AST, selection: Selection): t.Transformed {
         ...declarations
           .filter(isDeclarationInitialized)
           .map(function ({ id, init }) {
-            if ("typeAnnotation" in id) {
-              id.typeAnnotation = null;
+            if (id.type == "Identifier" && "typeAnnotation" in id) {
+              id = t.identifier(id.name);
             }
             return t.expressionStatement(t.assignmentExpression("=", id, init));
           })
--- a/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
+++ b/src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
@@ -34,7 +34,7 @@ function updateCode(ast: t.AST, selection: Selection): t.Transformed {
           .filter(isDeclarationInitialized)
           .map(function ({ id, init }) {
             if (id.type == "Identifier" && "typeAnnotation" in id) {
-              id = t.identifier(id.name);
+              id = t.identifier(id.name); // new identifier without annotation
             }
             return t.expressionStatement(t.assignmentExpression("=", id, init));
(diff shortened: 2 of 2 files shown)


src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.test.ts
```diff
@@ -60,6 +60,13 @@ lastName = "Doe";`
         expected: `let firstName, lastName, age;
 firstName = "Jane";
 lastName = "Doe";`
+      },
+      {
+        description: "declarations with type annotations",
```
nicoespeon, line 65: 👍  Thanks for adding a test to illustrate

src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
```diff
@@ -32,9 +32,12 @@ function updateCode(ast: t.AST, selection: Selection): t.Transformed {
         ),
         ...declarations
           .filter(isDeclarationInitialized)
-          .map(({ id, init }) =>
-            t.expressionStatement(t.assignmentExpression("=", id, init))
-          )
+          .map(function ({ id, init }) {
+            if (id.type == "Identifier" && "typeAnnotation" in id) {
```
nicoespeon, line 36: I think we can write:

```ts
if (t.isIdentifier(id)) {
```

And that should work. It will be less dependent of the implementation details.

nicoespeon (approved):
Awesome, thanks for the fix!

I'll update the Changelog & get that release this week with the other changes ❤️
I'll try to get some improvements I've been working on done too, then I'll release.

nicoespeon:
@allcontributors please add iulspop for bug

src/refactorings/split-declaration-and-initialization/split-declaration-and-initialization.ts
iulspop: It looks like you merged into master without adding your suggestion. Did you change your mind?

nicoespeon: @iulspop nope, I did it faster. It wasn't blocking so simpler to get this merged then refactor on master 👍 

https://github.com/nicoespeon/abracadabra/commit/9b6624b1e682ffbcbf9cd43ca36e43cfbf275059

PR #258, opened 2021-03-07
Improve error messages that come from Babel

### LINKED ISSUE #145: Improve error messages that come from Babel
## Describe what's missing

If the source code as a syntax error, Babel can't parse it and triggers an error.

Problem is: it's not clear for the end user _why_ the refactoring failed in this case, and how it could be fixed.

Here's a recent example from [the extension reviews](https://marketplace.visualstudio.com/items?itemName=nicoespeon.abracadabra&ssr=false#review-details):

> Great plugin, but it somtimes doesn't work. E.g. I tried to "convert if/else to switch" but got an error: "I'm sorry, something went wrong: Cannot read property 'buildError' of undefined"

## Expected behavior

If Babel fails at parsing the AST, the error message is clearer like:

> I'm sorry, I can't build the AST from the source code. This may be due to a syntax error that you can fix. Here's what went wrong: Cannot read property 'buildError' of undefined

## How it would be implemented

Here's the function that parse code into AST: https://github.com/nicoespeon/abracadabra/blob/558a8115ccc489f8dbb820d78a491e301cb6d148/src/ast/transformation.ts#L101

We could probably wrap this with a try/catch so we can enrich the error message that is thrown.

chrstnbrn:
Closes #145 

image

### DIFF
--- a/src/ast/transformation.test.ts
+++ b/src/ast/transformation.test.ts
@@ -1,5 +1,5 @@
 import * as t from "./domain";
-import { transform } from "./transformation";
+import { parse, transform } from "./transformation";
 
 describe("Transformation", () => {
   describe("transform", () => {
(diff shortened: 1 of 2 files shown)


nicoespeon:
@all-contributors please add chrstnbrn for code

PR #263, opened 2021-03-23
Add For...of -> forEach refactoring for arrays

### LINKED ISSUE #92: Make "Convert for to forEach" handle for…of loops
The for loop to foreach refactoring apparently does not work on `for (const entry of entries)` style loops. At least the plugin says it did not detect a valid for-loop.
Would be great to make it work for this style of loops.

Abracadabra version: 3.2.3

ZakMiller:
Resolves #92 

I'm new to the code base and AST/babel so let me know if there are some things you'd like me to change.

### DIFF
--- a/src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
+++ b/src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
@@ -1,8 +1,13 @@
 import { ErrorReason, Code } from "../../editor/editor";
 import { InMemoryEditor } from "../../editor/adapters/in-memory-editor";
 import { testEach } from "../../tests-helpers";
+import * as t from "../../ast";
 
-import { convertForToForeach } from "./convert-for-to-foreach";
+import {
+  canConvertForLoop,
+  convertForToForeach
+} from "./convert-for-to-foreach";
+import { Selection } from "../../editor/selection";
 
 describe("Convert For To Foreach", () => {
   testEach<{ code: Code; expected: Code }>(
@@ -310,6 +315,26 @@ for (let entry of typedArray) {
       expect(editor.code).toBe(originalCode);
     }
   );
+
+  it("should call onMatch for a matching for-loop", async () => {
+    const code = `for (let i = 0; i < items.length; i++) {
+  console.log(items[i]);
+}`;
+    const onMatch = jest.fn();
+
+    t.traverseAST(t.parse(code), {
+      enter(path) {
+        const visitor = canConvertForLoop(Selection.cursorAt(0, 0), onMatch);
+        const visitorNode = visitor[path.node.type];
+        if (typeof visitorNode === "function") {
+          // @ts-expect-error visitor can expect `NodePath<File>` but `path` is typed as `NodePath<Node>`. It should be OK at runtime.
+          visitorNode.bind(visitor)(path, path.state);
+        }
+      }
+    });
+
+    expect(onMatch).toHaveBeenCalled();
+  });
 
   it("should show an error message if refactoring can't be made", async () => {
     const code = `// This is a comment, can't be refactored`;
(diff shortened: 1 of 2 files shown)


nicoespeon:
Thank you @ZakMiller, I'll have a look in the next days 👍 ❤️

nicoespeon (approved):
@ZakMiller I want to thank you for this great PR 👏 

What I really liked:
- You added tests to cover all of your changes, proving they were working
- You extracted named functions that made the overall intent easy to follow

I did a couple of changes to polish a few things (I figured it was better than asking you to do the changes 10 days after you open the PR). These were:
- Added a test to verify that only the selected for-of would be converted (it was not the case, the fix was simple though => https://github.com/nicoespeon/abracadabra/pull/263/commits/b170c15a433a8b38a17f6bb4bfa156a541de0906). 
- `createVisitor()` should use the `onMatch()` callback that is provided. This is because [it's called by the `ActionProvider`](https://github.com/nicoespeon/abracadabra/blob/ec6b669b3ab37d6e3003799c57fad4cb66708f2a/src/action-providers.ts#L75-L86) VSCode uses to determine if the refactoring can be made. Concretely, this is what makes the refactoring appear in the quick fix suggestions (the lightbulb 💡 ). The way I solved it was by extracting what differed between the 2 `onMatch` implementations you created. I also realized this part wasn't covered with a test, which is why you missed it (I thought types would complain, but they won't complain if the argument isn't used at all, which is fair). I added that test.

I compiled the result to try that and it worked like a charm! I'll get this release by the end of the week.

Thank you again 👍

src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
```diff
@@ -149,6 +154,86 @@ items.forEach(item => {
 }`,
         expected: `this.data[0].items.forEach(item => {
   console.log(item);
+});`
+      },
+      {
+        description: "for-of",
+        code: `const items = ['foo', 'bar', 'baz'];
+
+[cursor]for (const val of items) {
```
nicoespeon, line 163: By default, the cursor is at the top of the code snippet. It looked suspicious then, because you wouldn't expect the refactoring to work on the for-of while the cursor isn't on it exactly.

That change allowed me to identify we forgot to check the for-of was indeed selected before refactoring it.

src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
```diff
@@ -180,6 +265,44 @@ items.forEach(item => {
...
+  console.log(character);
+}`
+      },
+      {
+        description: "for-of but an inline string",
+        code: `for (let character of 'abcde') {
+  console.log(character);
+}`
+      },
+      // 👇 These patterns could be converted (they support forEach) but are not implemented yet.
```
nicoespeon, line 283: I went for this, I don't like having TODO comments that aren't tied to a test. 

That's a nice comment to keep around, so we remember why features aren't supported. I won't create these issues yet, I'll rather wait for demand.

src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
```diff
@@ -193,6 +316,26 @@ items.forEach(item => {
     }
   );
 
+  it("should call onMatch for a matching for-loop", async () => {
```
nicoespeon, line 319: Not happy that this test is quite complex in its implementation, but nothing else is preventing us from forgetting to use the `onMatch()` argument.

I'll figure out something better after I did some more I guess.

nicoespeon:
@all-contributors please add @ZakMiller for bug, code

src/refactorings/convert-for-to-foreach/convert-for-to-foreach.test.ts
ZakMiller: Makes sense!

PR #375, opened 2021-08-28
Upgrade babel

nicoespeon:
I finally took the time to upgrade babel from 7.9 to the latest version (7.15), resolving the type conflicts.

A few code changes were required too. I'm glad we have a good test coverage: now I feel confident there's no regression (unless we are missing tests for these).

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "deploy:ovsx": "ovsx publish"
   },
   "devDependencies": {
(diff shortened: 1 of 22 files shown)


nicoespeon:
There should be no noticeable change for end-users.

But 2 benefits already:
- types are a little better/more accurate
- performance is better (the full tests suite runs in ~4s instead of ~20s on my machine)

PR #395, opened 2021-09-24
Implement toggle braces for loops

### LINKED ISSUE #247: Toggle Braces: handle `for` loops & `while`

There exist refactorings to add or remove braces for if/else or arrow functions, but not for single line loops (`for`, `while`, ...).

Behaving exactly as for if/else, I would like refactorings to add or remove braces from single line loops.

## Additional context

I don't know about your internals, but maybe it would make sense to group all add/remove braces refactorings as a general refactoring that is applicable wherever braces can be added or removed.

chrstnbrn:
Closes #247

Toggle braces now also works for
- for statements
- for in statements
- for of statements
- while statements
- do while statements

refactoring

I used the alias `Loop` to implement this refactoring. This way we don't have to handle all the different kinds of loops separately. However, I also had to change the action provider to be able to handle aliases. Without this the toggle braces refactoring was not suggested for loops.

### DIFF
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -33,10 +33,9 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     try {
       const ast = t.parse(editor.code);
 
-      return this.findApplicableRefactorings(
-        ast,
-        editor.selection
-      ).map((refactoring) => this.buildCodeActionFor(refactoring));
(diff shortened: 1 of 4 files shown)


nicoespeon (approved):
👏 I'm very impressed with the quality of your changes. Thanks for adding the relevant tests and considering the different scenarios that could happen.

Looks all good to me. I'll get it shipped in the next release later this week 🎉

PR #417, opened 2021-10-05
[New Refactoring] forEach to for-of

ianobermiller:
New refactoring to convert `array.forEach` into `for (const item of array)`. Also converts `return` statements into `continue` statements.

for-each-to-for-of


nicoespeon:
@ianobermiller awesome! I'll review that tomorrow, but I'm pleased to see you wrote tests and even recorded a gif of the refactoring in action ❤️ 

Thanks!

PR #459, opened 2021-12-19
migrate tslint to eslint

tjx666:
1. migrate tslint to eslint
2. fix eslint lint errors
3. replace pretty-quick with lint-staged
4. fix some ts compile error

But I occured an issue that I can't activate extenison. This issue is related with module `code-block-writer` which is dependent by `ts-morph`。

My version info:

```
Version: 1.64.0-insider (Universal)
Commit: d12df34e31b4a018735d312a8947d79331132368
Date: 2021-12-17T05:14:21.805Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 21.2.0
```

image


tjx666:
This project compile costs too long. One reason maybe whenever this ts project contains all test files. If I have free time, often weekend, I may help to do some more optimization to this project.

nicoespeon:
Thanks for helping @tjx666. I will have a look at that too. I noticed that error recently, and I'm not sure where this is coming from. The file does exist, so that's weird.

cc @fabien0102

src/playground/gilded-rose.ts
```diff
@@ -20,19 +20,19 @@ export class GildedRose {
   updateQuality() {
     for (let i = 0; i < this.items.length; i++) {
       if (
-        this.items[i].name != "Aged Brie" &&
-        this.items[i].name != "Backstage passes to a TAFKAL80ETC concert"
+        this.items[i].name !== "Aged Brie" &&
```
nicoespeon, line 23: Oh, the playground is intentionally invalid. We should exclude this folder from the linter. 

The playground code is the one that will open in the Extension Host when you debug locally.

src/refactorings/extract-generic-type/extract-generic-type.ts
```diff
@@ -40,7 +40,7 @@ async function extractGenericType(editor: Editor) {
 
 function findAllOccurrences(ast: t.AST, selection: Selection): AllOccurrences {
   let selectedOccurrence: Occurrence | null = null;
-  let otherOccurrences: Occurrence[] = [];
+  const otherOccurrences: Occurrence[] = [];
```
nicoespeon, line 43: 👍 Nice, thanks for fixing all of these

src/refactorings/split-multiple-declarations/split-multiple-declarations.ts
```diff
@@ -36,13 +36,11 @@ function updateCode(ast: t.AST, selection: Selection): t.Transformed {
 }
 
 function createVisitor(
-  selection: Selection,
+  _selection: Selection,
   onMatch: (path: t.NodePath<t.VariableDeclaration>) => void
 ): t.Visitor {
   return {
     VariableDeclaration(path) {
-      selection;
```
nicoespeon, line 44: 👍 Good catch.

src/refactorings/rename-symbol/rename-symbol.ts
```diff
@@ -13,7 +13,7 @@ async function renameSymbol(editor: Editor) {
   }
 }
 
-function findSelectedSymbol(editor: Editor): Symbol {
+function findSelectedSymbol(editor: Editor): Nothing {
```
nicoespeon, line 16: That doesn't seem right. `result` may be an `Identifier`. Both are implementing the `Symbol` interface, so the previous code should have been correct. What was the linter issue?

src/type-checker/type-checker.ts
```diff
@@ -108,8 +108,7 @@ export class TypeChecker {
     program: ts.Program
   ): ts.Node | undefined {
     try {
-      // @ts-ignore Internal method
-      return ts.getTouchingPropertyName(
+      return (ts as any).getTouchingPropertyName(
```
nicoespeon, line 111: Hmm I'd rather keep a `@ts-expect-error` here, explaining it's an internal method we are using. I don't like parsing stuff to `as any` because it bypasses type checking. `@ts-expect-error` leaves the door open for the problem to be solved in the future.

package.json
```diff
@@ -65,21 +67,24 @@
...
     "@vscode/test-electron": "1.6.2",
     "all-contributors-cli": "6.20.0",
     "chai": "4.3.4",
+    "eslint": "^8.5.0",
+    "eslint-config-prettier": "^8.3.0",
     "glob": "7.2.0",
     "husky": "7.0.4",
     "hygen": "6.1.0",
     "jest": "27.4.2",
+    "lint-staged": "^12.1.3",
```
nicoespeon, line 81: 🍰 I prefer to omit the caret and install exact versions of packages. That leaves no room for surprises. The JS world isn't very SemVer friendly, so I had surprises in the past with flexible versions being installed.

We have good tests and dependabot regularly makes sure we keep up-to-date, so that's not an issue.

.eslintrc.js
```diff
@@ -0,0 +1,47 @@
...
+    "new-cap": ERROR,
+    "no-invalid-this": ERROR,
+    "no-unused-expressions": ERROR,
+
+    "consistent-this": OFF,
+    "no-control-regex": OFF,
+    "no-empty": OFF,
+    "no-use-before-define": OFF
+  }
+};
```
nicoespeon, line 47: 👍 Nice!

src/refactorings/inline/inline-variable/find-inlinable-code.ts
```diff
@@ -248,7 +249,7 @@ class InlinableIdentifier implements InlinableCode {
         let selection = Selection.fromAST(loc);
 
         if (t.isTemplateLiteral(parent)) {
-          code = code.replace(/^("|'|\`)/, "").replace(/("|'|\`)$/, "");
```
nicoespeon, line 251: That broke the tests. It should really be `

Was it changed automatically? If so, we should fix some config.

nicoespeon:
@tjx666 can you fix the conflicts and address the few comments I left?

On my side, I'll investigate to fix the code-block-writer issue when activating the extension + try to make the project compile faster.

tjx666:
Fix them this weekend

nicoespeon:
@tjx666 FYI, the "Cannot find module './comment-char.js'" was a bug caused by the upgrade of ts-morph to v13. I fixed that on the main branch (commit 43ed32f8db0c91ecac86f00edfd03b84c1915e53). 

That doesn't optimize the perfs, but at least it works now 👍

tjx666:
@nicoespeon fixed

nicoespeon:
@tjx666 working like a charm! Thank you for this super cool upgrade ❤️

nicoespeon:
@all-contributors please add tjx666 for infra

PR #478, opened 2022-01-07
fix: exit visitor

j4k0xb:
Fixes the quick action provider
`forEachToForOf` used an exit method instead of the shorthand form (babel handbook), so I added the calls
> Note: `Identifier() { ... }` is shorthand for `Identifier: { enter() { ... } }`.

Before:
before

After:
after

-----------------------

I had some trouble with getting the current version to build so I tested it on v6.8.0 (3d7cc05a06ad4015313374bb27455c7ace7e0cf6) for now
error

### DIFF
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -104,9 +104,17 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     }
 
     const visitorNode = this.getVisitorNode(visitor, path);
+    // call enter shorthand of e.g. { Identifier() { ... } }
     if (typeof visitorNode === "function") {
       // @ts-expect-error visitor can expect `NodePath<File>` but `path` is typed as `NodePath<Node>`. It should be OK at runtime.
       visitorNode.bind(visitor)(path, path.state);
+    } else if (typeof visitorNode === "object" && visitorNode !== null) {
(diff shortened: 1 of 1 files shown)


src/action-providers.ts
```diff
@@ -104,9 +104,17 @@ class RefactoringActionProvider implements vscode.CodeActionProvider {
     }
 
     const visitorNode = this.getVisitorNode(visitor, path);
+    // call enter shorthand of e.g. { Identifier() { ... } }
     if (typeof visitorNode === "function") {
       // @ts-expect-error visitor can expect `NodePath<File>` but `path` is typed as `NodePath<Node>`. It should be OK at runtime.
       visitorNode.bind(visitor)(path, path.state);
+    } else if (typeof visitorNode === "object" && visitorNode !== null) {
+      // call methods of e.g. { Identifier: { exit() { ... } } }
```
nicoespeon, line 112: 👍 Good catch! Thanks for adding the comment to clarify.

nicoespeon (approved):

nicoespeon:
Oh oh, nice one @j4k0xb. You rock ❤️ 

P.S: I figured out the issue with the extension not compiling. Turns out it's due to the ts-morph upgrade to v13 that doesn't work well. I downgraded it back to v12 until this gets fixed (we don't really need the latest version TBH).

PR #480, opened 2022-01-09
ignore docs dir with gifs to reduce extension size

zardoy:
I just noticed that installing this extension takes forever. Published vsix is 40mb. After ignoring docs with gifs `vsce` says it is 5mb now.

ImageOptim said it is possible to save 11mb in repo. ~~However GIFs are evil and I suggest to remove them from repo and use fancy codes instead.~~ UPD: I understand there is no other way to demonstrate functionality, however I believe in some future we can use SVGs for that.

You can also [reduce bundle size](https://github.com/prettier/prettier-vscode/pull/2302#issuecomment-1003753084) with esbuild. It can also give you really fast builds!
WDYT?

Also note that it is much simpler to ignore everything in `.vscodeignore` and unignore only needed assets.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,7 +1,7 @@
 .github/**
 .vscode/**
 dist/**
-docs/*
+docs/**
 !docs/logo
 /docs/logo/*
 !docs/logo/abracadabra-vignette.png
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 out
 node_modules
+.DS_Store
 .vscode-test/
 .vscode-test-web/
 *.vsix


.vscodeignore
```diff
@@ -1,7 +1,7 @@
 .github/**
 .vscode/**
 dist/**
-docs/*
+docs/**
```
nicoespeon, line 4: ❤️

nicoespeon (approved):
Love it, a great catch! I didn't check the resulting .vsix in a long time, so I didn't realize the gifs were bundled (which is stupid).

Thank you for this, you rock 👍

nicoespeon:
I'm also open for contributions to improve bundle size and time. I'd just be mindful of #459 (which I'm about to review and merge) to make sure you don't have conflicts when doing so.

nicoespeon:
@all-contributors please add zardoy for code and infra

nicoespeon:
I'm releasing this improvement (v6.9.1) so people save time & data when installing the extension 😉

Issue #568, opened 2022-04-19
Convert Switch case with only a default case to if-statement


11joselu:
## Describe what's missing

Convert Switch case with only a **default case**  to if-statement

## How would it work

Given this input
```javascript
 switch (true) {
    default:
        myVar = 'Hello';
        break;
    }
```

## Expected behavior

The output should be

```javascript
if (true) {
  myVar = 'Hello';
 }
```

## Screenshots

Right now the extension is not able to detect this case

![image](https://user-images.githubusercontent.com/8685132/164041489-0d5a93e8-3968-4a3f-a027-2c52ea7554f1.png)

## Additional information

- Version of the extension used: v6.12.0

How we arrive into a switch with only a default case?

[Doing Tennis Refactoring Kata](https://github.com/emilybache/Tennis-Refactoring-Kata/blob/main/javascript-jest/tennis1.js) After some refactor I endend at [L34](https://github.com/emilybache/Tennis-Refactoring-Kata/blob/main/javascript-jest/tennis1.js#L34) with

```javascript
switch (tempScore) {
      default:
        score += scores[tempScore];
        break;
    }
```

To simplify more the code. I can convert that switch to if an then apply "Remove Dead Code".
![vscode-abracadabra](https://user-images.githubusercontent.com/8685132/164043105-ea5c8d5f-a16d-4746-87a3-092c40c4ad5f.gif)

nicoespeon:
Excellent suggestion and thanks for the detailed report @11joselu 👏 
The Tennis refactoring kata is a great one. Hopefully, it will be even faster to do with Abracadabra now 😉 

@all-contributors please add @11joselu for ideas

PR #570, opened 2022-04-21
Convert Switch case with only a default case to if-statement

11joselu:
Before we are not able to convert switch-to.if when there are not cases. Now you can convert a default switch to if statement then remove dead code. Related to https://github.com/nicoespeon/abracadabra/issues/568

vscode-new-switch

What an amazing Developer experiencie!!! Congrats :)


nicoespeon:
Thanks @11joselu, I missed your PR and implemented it on my side. The implementation is slightly different, with an extra test: https://github.com/nicoespeon/abracadabra/commit/c8a7b519a49264c361d729bc5cd8de009661ff37

Closing this PR. I'll release the feature today.

PR #593, opened 2022-05-11
[New Refactoring] Extract useCallback

ianobermiller:
Add a new React-specific refactoring to extract an inline function from a JSX prop to a `useCallback` call.

## Handled
- Arrow and function expressions
- Rename the extracted identifier if it conflicts
- Add a block if it was an arrow expression without braces

## Not Handled
- Add types to the function
- Infer dependencies (always uses an empty array) - use the React hooks ESLint fixers for that

https://user-images.githubusercontent.com/897931/167899020-6e456b0c-6c61-4b02-abdf-449cb9ce8a68.mp4

## Test Plan

- Unit tests
- "Run in debugger" and test on a real React file

### DIFF
--- a/src/refactorings/react/extract-use-callback/extract-use-callback.test.ts
+++ b/src/refactorings/react/extract-use-callback/extract-use-callback.test.ts
@@ -1,0 +1,89 @@
+import { InMemoryEditor } from "../../../editor/adapters/in-memory-editor";
+import { Code, ErrorReason } from "../../../editor/editor";
+import { testEach } from "../../../tests-helpers";
+import { extractUseCallback } from "./extract-use-callback";
+
+describe("Extract useCallback", () => {
+  testEach<{ code: Code; expected: Code }>(
+    "should extract useCallback from an inline function",
+    [
+      {
+        description: "arrow expression",
+        code: `
+function Bar() {
+  return <Foo onFoo={[cursor](e) => console.log(e)} />;
+}`,
+        expected: `
+function Bar() {
+  const onFoo = useCallback(e => console.log(e), []);
+  return <Foo onFoo={onFoo} />;
+}`
+      },
+      {
+        description: "function expression",
+        code: `
+function Bar() {
+  return (
+    <Foo onFoo={[cursor]function handleFoo() {
+      console.log(e);
+    }} />
+  );
+}`,
+        expected: `
+function Bar() {
+  const onFoo = useCallback(function handleFoo() {
+    console.log(e);
+  }, []);
+
+  return <Foo onFoo={onFoo} />;
+}`
+      },
+      {
+        description: "uses a unique name",
+        code: `
+function Bar({onFoo}) {
+  return <Foo onFoo={[cursor](e) => {console.log(e); onFoo(e); }} />;
+}`,
+        expected: `
+function Bar({onFoo}) {
+  const onFoo2 = useCallback(e => {console.log(e); onFoo(e); }, []);
+  return <Foo onFoo={onFoo2} />;
+}`
+      },
+      {
+        description: "adds a block if necessary",
+        code: `
+const Bar = () => <Foo onFoo={[cursor](e) => {console.log(e); onFoo(e); }} />;`,
+        expected: `
+const Bar = () => {
+  const onFoo = useCallback(e => {console.log(e); onFoo(e); }, []);
+  return <Foo onFoo={onFoo} />;
+};`
+      }
+    ].map(({ description, code, expected }) => ({
+      description,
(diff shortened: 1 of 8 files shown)


nicoespeon:
Thanks @ianobermiller, that's awesome! I'll dig into this in more details later, but I'm impressed with the quality of your PR 👍 

For the dependencies, we already are using ESLint internally. I made a POC with @fabien0102 to see if we could use the autofix feature directly for the user, and this is working:

```ts
import { ESLint } from "eslint";

const eslint = new ESLint({
  fix: true,
  useEslintrc: false,
  overrideConfig: {
    plugins: ["react-hooks"],
    rules: {
      "react-hooks/exhaustive-deps": "error"
    },
    parserOptions: {
      ecmaFeatures: {
        jsx: true
      },
      ecmaVersion: 6,
      sourceType: "module"
    }
  }
});

describe("Try auto-fix", () => {
  it("should work", async () => {
    const code = `
function MyComponent() {
  const uploading = someFun();
  const uploading2 = someFun();
  const triggerFileBrowser = useCallback(() => {
    if (uploading) {
      return
    }
    if (uploading2) {
      return
    }

    fileInputRef.current.click()
  }, [])
}`;
    const results = await eslint.lintText(code);

    const NO_FIX = {
      fix: { range: [0, 0], text: "" }
    };
    const { fix } = results[0]?.messages[0]?.suggestions?.[0] ?? NO_FIX;
    const result =
      code.slice(0, fix.range[0]) + fix.text + code.slice(fix.range[1]);

    expect(result).toBe(`
function MyComponent() {
  const uploading = someFun();
  const uploading2 = someFun();
  const triggerFileBrowser = useCallback(() => {
    if (uploading) {
      return
    }
    if (uploading2) {
      return
    }

    fileInputRef.current.click()
  }, [uploading, uploading2])
}`);
  });
});
```

It requires installing `eslint-plugin-react-hooks` as a dependency, and probably declaring "eslint" as a production dependency too.

Could you try to use that snippet so the refactoring does handle the dependencies too? The idea would be to run the autofix on the produced code, before we write it back into the editor.

src/refactorings/react/extract-use-callback/extract-use-callback.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      },
+      {
+        description: "uses a unique name",
+        code: `
+function Bar({onFoo}) {
+  return <Foo onFoo={[cursor](e) => {console.log(e); onFoo(e); }} />;
+}`,
+        expected: `
+function Bar({onFoo}) {
+  const onFoo2 = useCallback(e => {console.log(e); onFoo(e); }, []);
```
nicoespeon, line 49: 👏 for handling conflicting names like that!

And +1 for not iterating the counter for more than 10. That will be very, very, _very_ unlikely.

src/refactorings/react/extract-use-callback/extract-use-callback.ts
```diff
@@ -0,0 +1,84 @@
...
+        counter++;
+
+        // Avoid an infinite loop, just in case
+        if (counter > 10) {
+          return;
+        }
+      }
+
+      const id = t.identifier(extractedName);
+      const node = template.statement`
```
nicoespeon, line 40: TIL this exists. Nice 👏

nicoespeon (approved):

ianobermiller:
That is a really cool idea, I'll take a look at integrating ESLint.

nicoespeon (approved):
cheers-mate

This PR is awesome @ianobermiller. Kudos 👏

nicoespeon:
@all-contributors please add @ianobermiller for code and idea

nicoespeon:
Just a heads up: it's not merged yet because I realized that it doesn't work when we run it as an extension (eg. Debug the extension). The culprit is ESLint, and that's because it fails to load the parser & plugins. 

I investigated this with @fabien0102 and it turns out that the way eslint tries to load the parser/plugins doesn't work once everything is bundled. I think it would just work if we could pass the parser/plugins reference directly to ESLint instead of relying on a magic string. Or if it would just use a global `require()` instead of trying to resolve based on the `cwd` like it does (and which doesn't work once bundled).

Anyway, I'm working on fixing that. I really want the ESLint auto-fix, it's super valuable. I'll probably fork the eslint project to have a way to inject the direct reference to the parser & plugins instead of a magic string. I'll only use that fork as a dedicated package, for this specific feature. So that should work fine without too much extra maintenance burden.

ianobermiller:
Bummer! Hope you can get it working with a fork, or even get it upstreamed.

PR #664, opened 2022-07-25
Fix typo: versionning -> versioning

ash211:


nicoespeon:
Thank you @ash211, good catch 👍

nicoespeon:
@all-contributors please add ash211 for docs

PR #706, opened 2022-09-23
Enhance the default name for new expression

azhiv:
When extracting a class instantiation, for instance `new ImageEditor()`, provide a more user-friendly default name (instead of `extracted`). The new default is the camel-cased class name, `imageEditor` in the example above.


nicoespeon (approved):
Oooooh, that's a simple, yet awesome one! Thanks for updating the tests. It's an easy one to review, looking good 👍

nicoespeon:
@all-contributors please add azhiv for ideas and code

azhiv:
Thank you, @nicoespeon!

Issue #341, opened 2021-07-14
Re-order function arguments


qpwo:
## Is this request related to a problem? Please describe.

I'm always frustrated when I want to change the order of arguments to a function and I have to manually go through the callers and update them. I usually make a mistake and introduce a bug.

## Describe the solution you'd like

Select a function, abracadabra re-order arguments. Something like dragging boxes would be nifty but it's fine if I have to fill out a textbox with the same names in a different order. Then the function

Example: before:

```js
// in utils.js
function repeat(f, n) {
    return Array(n).fill(null).map((_,i) => f(i))
}
// in app.js
const squares = repeat(i => i*i, 5)
```

After:

```js
// in utils.js
function repeat(n, f) {
    return Array(n).fill(null).map((_,i) => f(i))
}
// in app.js
const squares = repeat(5, i => i*i)
```

nicoespeon:
That would be useful indeed. The difficult part is to get the exported references (it has to work cross-files), in particular with JavaScript code.

I believe this may be easier to address once https://github.com/nicoespeon/abracadabra/issues/275 is tackled. Using `typescript` at the project level would give us the AST cross-files (today refactorings are limited within the same file).

PR #725, opened 2022-10-09
[Proposal] Change function signature

11joselu:
Related to https://github.com/nicoespeon/abracadabra/issues/341

Hi @nicoespeon 

Here is a new refactoring proposal: Change function signature.

At the moment we are able to change "parameters arguments" of a function.

Cases:

- Given a function with two parameters:

```javascript
function add(param: number, num2: number) {
  return param + num2;
}

add(1, 200);
```
- Expect to change parameters and arguments orders

```javascript
function add(num2: number, param: number) {
  return param + num2;
}

add(200, 1);
``` 

function-arguments-change

> Question: Why "Change Signature" open a Tab for order definition?
> Answer: I not able to create a "Popup" for asking orders. vscode does not contain a (at the momento) API to give us that functionality. So I render a "Webview" as a sidebar. I found that [Color Picker](https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color) has a Popup but I don't want to look how they created it (lazy boy :P)

- Given a function with second param as optional
```javascript
function add(paramRequired, paramOptional: number = 1) {
  return paramRequired + paramOptional;
}

add(200);
```
- Expect to put `undefined` as a value when there are no value in call expression

```javascript
function add(paramOptional: number = 1, paramRequired) {
  return paramRequired + paramOptional;
}

add(undefined, 200);
```

There are more cases in test case.

- Given a function with value and callback
```javascript
function repeat(f, n) {
  return Array(n)
    .fill(null)
    .map((_, i) => f(i));
}

const squares = repeat((i) => i * i, 5);
```
- Expect to change argument order even if there a callback
```javascript
function repeat(n, f) {
  return Array(n)
    .fill(null)
    .map((_, i) => f(i));
}

const squares = repeat(5, (i) => i * i);
```

### What about modules

I develop the funcionality to search (using vscode Find References) references in base code and change arguments orders based on user definition

modules

### To Fix

For modules imported with aliases would not found a references and the changes will break app -> need to look how to prevent changes when there are aliases for a exported funcion

### Next?

Maybe we can add functionalities like:
- Remove param
- Add new param with a value or default value
Feedback are welcome :)


nicoespeon:
Hi @11joselu and thanks for this! I need to take some time to review the feature, UX, and implementation. I will get back to you as soon as possible on this 👍 

Overall, it's exciting. "Change Function Signature" is an ambitious, yet powerful refactoring.

src/editor/adapters/in-memory-editor.ts
```diff
@@ -59,8 +63,8 @@ export class InMemoryEditor implements Editor {
     return Promise.resolve();
   }
 
-  async writeIn(path: RelativePath, code: Code): Promise<void> {
-    this.otherFiles.set(path, new InMemoryEditor(code));
+  async writeIn(path: Path, code: Code): Promise<void> {
+    this.otherFiles.set(path as RelativePath, new InMemoryEditor(code));
```
nicoespeon, line 67: Is it necessary to force the cast here? I'll have a look.

src/editor/adapters/vscode-editor.ts
```diff
@@ -182,6 +193,83 @@ export class VSCodeEditor implements Editor {
...
   }
+
+  async getSelectionReferences(selection: Selection): Promise<CodeReference[]> {
+    const locations = (await vscode.commands.executeCommand(
+      "vscode.executeReferenceProvider",
+      this.document.uri,
+      selection.start
+    )) as vscode.Location[];
+
+    const references = locations.map((loc) => {
```
nicoespeon, line 204: 🍰 We may inline this temp variable and directly return the `locations.map()`

src/editor/adapters/vscode-editor.ts
```diff
@@ -213,3 +301,161 @@ function toVSCodeCommand(command: Command): string {
...
+          <td class="params-name">${name}</td>
+          <td>
+            <span class="up"></span>
+            <span class="down"></span>
+          </td>
+        </tr>
+    `;
+  });
+
+  const html = `
```
nicoespeon, line 322: 🍰 Ideally, we would move most of this into an actual HTML file to get syntax highlighting and all. But this is good enough here (especially since it makes it more convenient to inject the dynamic content).

src/playground/maths/calculator.ts
```diff
@@ -0,0 +1,11 @@
...
+
+add(3, 400);
+
+// Used to view what params generates in webview table
+function viewParametersTable(
+  literal: number | string | boolean,
+  obj = {},
+  [desc, descTwo, ...rest]: any[],
+  { item, itemTwo, ...anotherRest }: Record<string, any>
+) {}
```
nicoespeon, line 11: 👍 for adding playground files

src/editor/adapters/in-memory-editor.ts
```diff
@@ -204,6 +208,87 @@ export class InMemoryEditor implements Editor {
   removeLine(line: number): void {
     this.codeMatrix.splice(line, 1);
   }
+
+  async getSelectionReferences(selection: Selection): Promise<CodeReference[]> {
```
nicoespeon, line 212: 🍰 Ideally, I would like to write one or two contract tests to verify that this behavior is similar to what VS Code actually does. I don't write a lot of these tests since they are slower and more likely to fail, but just a handful is enough to cover the blind spots introduced with the In-Memory implementation.

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -0,0 +1,404 @@
...
+import { testEach } from "../../tests-helpers";
+
+import { changeSignature } from "./change-signature";
+
+type TestSample = {
+  code: Code;
+  path: RelativePath;
+};
+
+describe("Change Signature", () => {
```
nicoespeon, line 12: 👍 Kudos for adding tests that document the different scenarios this is already handling. Love it!

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  }>(
+    "In same file",
+    [
+      {
+        description: "when there are a function without references",
+        setup: {
+          currentFile: `function [cursor]add(a, b) {
+            return a + b;
+          }`,
+          path: new RelativePath("./aFileWitoutReferences.ts")
```
nicoespeon, line 25: 🍰 For these tests, since everything is supposed to be in the same file, I think we can omit the `path` attribute altogether and just make the test body use the same path to write the code and read it for asserts. What do you think?

nicoespeon: So it would be a regular `code: string; expected: string` instead of a `setup/expected`

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -0,0 +1,404 @@
...
+      setup: {
+        currentFile: TestSample;
+        otherFiles: TestSample[];
+      };
+      expected: {
+        currentFile: TestSample;
+        otherFiles: TestSample[];
+      };
+    }>(
+      "should change signature of function with a reference in modules",
```
nicoespeon, line 246: 👌 Perfect

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  const promises = setup.otherFiles.map(async (file) => {
+    await editor.writeIn(file.path, file.code);
+  });
+
+  return Promise.all(promises);
+}
+
+function userChangePositionOf(startAt: number, endAt: number) {
+  return {
+    label: Date.now.toString(), // we don't care about param name in inMemory
```
nicoespeon, line 398: 🍰 In such a situation, I like to make it explicit and put: `label: 'irrelevant'` instead

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  editor: InMemoryEditor
+) {
+  const promises = setup.otherFiles.map(async (file) => {
+    await editor.writeIn(file.path, file.code);
+  });
+
+  return Promise.all(promises);
+}
+
+function userChangePositionOf(startAt: number, endAt: number) {
```
nicoespeon, line 396: 👍 Nice abstraction. It makes tests easier to read

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
+import { Editor, ErrorReason, SelectedPosition } from "../../editor/editor";
+import { Selection } from "../../editor/selection";
+import * as t from "../../ast";
+import { Path } from "../../editor/path";
+
+let toModifyNode: t.FunctionDeclaration;
+
+export async function changeSignature(editor: Editor) {
+  const params =
+    toModifyNode?.params.map((p, index) => {
```
nicoespeon, line 10: I'm intrigued by the fact we don't seem to parse the code first and we rely on the current value of `toModifyNode` that was set by `createVisitor`. I guess this works because the later is invoked regularly to verify if a change can be made (action provider). 

I need to check what happens if you attempt to execute the refactoring from the command palette instead of the lightbulb. I guess it would still work and it's more performant… but I _feel_ it would be safer to parse the code first to determine the node to modify. Hopefully, that wouldn't be a big change to make.

nicoespeon: I'm also concerned with the fact the variable persists and what could happen if you trigger the refactoring again from the command palette with an old (maybe invalid) value?

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
...
+      return {
+        label: getParamName(p),
+        value: {
+          startAt: index,
+          endAt: index
+        }
+      };
+    }) || [];
+
+  await editor.askForPositions(params, async (message) => {
```
nicoespeon, line 20: 🍰 `newPositions` may be a better name than `message` here.

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
...
+        return true;
+      })
+    );
+  });
+}
+
+function updateCode(
+  ast: t.AST,
+  selection: Selection,
+  orders: SelectedPosition[]
```
nicoespeon, line 80: 🍰 `newPositions` here too (for consistency)

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    await Promise.all(
+      result.map(async (result) => {
+        await editor.writeIn(
+          result.path,
+          alreadyTransformed[result.path.value]
+        );
+
+        return true;
```
nicoespeon, line 71: Is this return necessary? 🤔

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
...
+          message
+        );
+
+        alreadyTransformed[x.path.value] = `${transformed.code}`;
+
+        result.push({
+          path: x.path,
+          transformed
+        });
+      } catch (error) {
```
nicoespeon, line 58: 🍰 Could we log this error in the console at least? That could help debugging.

src/refactorings/change-signature/change-signature.ts
```diff
@@ -0,0 +1,198 @@
...
+    FunctionDeclaration(path) {
+      if (!selection.isInsidePath(path)) return;
+
+      toModifyNode = path.node;
+      onMatch(path);
+    }
+  };
+}
+
+function getParamName(
```
nicoespeon, line 133: 👍 Nice little helper here

nicoespeon:
@all-contributors please add 11joselu for code and ideas

src/editor/adapters/vscode-editor.ts
```diff
@@ -110,8 +114,15 @@ export class VSCodeEditor implements Editor {
     }
   }
 
-  protected fileUriAt(path: RelativePath): vscode.Uri {
-    const filePath = path.absoluteFrom(this.document.uri.path);
+  protected fileUriAt(path: Path): vscode.Uri {
+    let filePath;
+
+    if (path instanceof RelativePath) {
```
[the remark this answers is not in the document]
11joselu: I changed the code but I need you to check if I understood the changes correctly 🙏🏽 
https://github.com/nicoespeon/abracadabra/pull/725/commits/9439a47c2fe3a72cd72e971efcd6d39aa17d439c

src/editor/adapters/vscode-editor.ts
11joselu: Fixed :)

src/refactorings/change-signature/change-signature.ts
11joselu: Fixed at https://github.com/nicoespeon/abracadabra/pull/725/commits/d4d2cd1275c0724182d187e94de0dba685cd6c34

src/refactorings/change-signature/change-signature.ts
11joselu: Fixed at https://github.com/nicoespeon/abracadabra/pull/725/commits/d4d2cd1275c0724182d187e94de0dba685cd6c34

src/refactorings/change-signature/change-signature.ts
11joselu: Is not necessary. Nice catch :) 

Fixed at https://github.com/nicoespeon/abracadabra/pull/725/commits/628d049796d38400b6bc2c267c9445788aae4bb4

src/refactorings/change-signature/change-signature.ts
11joselu: console.log added in https://github.com/nicoespeon/abracadabra/pull/725/commits/bea756d2a38c293388e5ade61c39ed77c1dffdc1

src/refactorings/change-signature/change-signature.test.ts
11joselu: Make sense to me.

Changes done at https://github.com/nicoespeon/abracadabra/pull/725/commits/8eae0dbca6b41b0215134d6c65fc1735a53e6ac2

src/editor/adapters/in-memory-editor.ts
11joselu: With this https://github.com/nicoespeon/abracadabra/pull/725#discussion_r1009991380 changes, cast is not required anymore. 

Changes:
- Remove cast and fix typos https://github.com/nicoespeon/abracadabra/pull/725/commits/5695d2da50fa8f4e98090552ce59361bd4da43b4
-  Use AbsolutePath in test too https://github.com/nicoespeon/abracadabra/pull/725/commits/d2d8425a4370673ec502bdd74c5ace5026ac23e9

src/refactorings/change-signature/change-signature.test.ts
11joselu: Yeah!!
With this changes test cases are much simpler 

https://github.com/nicoespeon/abracadabra/pull/725/commits/05cf6cdf42fbd4ff3cd4c2ed92edb8968aee0961

src/editor/adapters/vscode-editor.ts
11joselu: Maybe, at the moment, we can keep inlined html and in the next iteration (_introduce param or remove param_) move that content into a html file ([something like this](https://stackoverflow.com/a/56698125)).

src/editor/adapters/vscode-editor.ts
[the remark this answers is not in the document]
nicoespeon: Yep, that's what I had in mind 👍

src/editor/adapters/vscode-editor.ts
nicoespeon: Yep, I think this is good enough for now. I forgot to mention it but when I prefix my comment with 🍰 it usually means it's a suggestion/idea/nice-to-have, but not blocking change.

All is good with this. Good job in addressing comments already. I'll take some time to play with the refactoring myself and likely merge that so it will be shipped in the next release 👍

nicoespeon:
It's all good!

I didn't see any error. I tried with regular JS files and it was able to resolve references just fine. IMO, that's good enough to ship. I'll document the feature and all.

Other ideas for future iterations @11joselu:
- Handle arrow functions (eg. `export const add = (x, y) => {}`
- Handle object/class methods
- Support renaming params from the webview

PR #728, opened 2022-10-13
feat: add svelte support

### LINKED ISSUE #726: Add support for .svelte files

The plugin does not offer suggestions in svelte files.

Refactoring suggestions to appear in svelte files. 

## Additional context

With my limited knowledge of the plugin / VSCode plugins in general. I feel like this would be quite similar to adding support for vue files, covered in https://github.com/nicoespeon/abracadabra/pull/157/files . 

Svelte uses the same convention as Vue for JS / Typescript with the code between the script tag. 

```html
<script lang="ts">
// code here
</script>
```

I'm not sure how far the code base has moved on from that PR, but if it is implemented in a similar way, hopefully it would be reasonably easy to add support for svelte? 

Please let me know if you think it would be more complicated for whatever reason.

srsholmes:
Fixes #726 

Adds support for .svelte files. 

I piggy backed off the support for vue files, which seems to work, using this PR as a base https://github.com/nicoespeon/abracadabra/pull/157/files. 

I wasn't sure if it was worth renaming the `VueVSCodeEditor` to something more generic like `VueAndSvelteVSCodeEditor`

I havn't added in any additional tests yet but can do if required. 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "javascript",
     "typescript",
     "react",
-    "vue"
+    "vue",
+    "svelte"
   ],
   "categories": [
     "Formatters",
@@ -165,7 +166,8 @@
     "onLanguage:javascriptreact",
     "onLanguage:typescript",
     "onLanguage:typescriptreact",
-    "onLanguage:vue"
+    "onLanguage:vue",
+    "onLanguage:svelte"
   ],
   "contributes": {
     "commands": [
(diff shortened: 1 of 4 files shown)


nicoespeon:
That's perfect @srsholmes 👍 

I am not familiar with Svelte ecosystem and `.svelte` files, but if it works like `.vue` ones then this is working! I tested it myself from your PR and it does work indeed.

Thanks for bringing it up and making the change. I will take care of renaming the file and all to be consistent.

nicoespeon:
@all-contributors please add srsholmes for ideas and code

PR #746, opened 2022-11-02
Change Signature I: Add contract test for code references

11joselu:
Verify that `getSelectionReferences` behavior is similar to what VS Code and InMemory does

It was pending from https://github.com/nicoespeon/abracadabra/pull/725#discussion_r1009994929

### DIFF
--- a/src/editor/editor-contract-test.ts
+++ b/src/editor/editor-contract-test.ts
@@ -2,9 +2,10 @@ import { suite, test, afterEach } from "mocha";
 import { assert } from "chai";
 import * as sinon from "sinon";
 
-import { Editor, Code, RelativePath } from "./editor";
+import { Editor, Code, RelativePath, AbsolutePath } from "./editor";
 import { Position } from "./position";
 import { Selection } from "./selection";
+import { CodeReference } from "./code-reference";
 
 /**
  * This is a contract tests factory.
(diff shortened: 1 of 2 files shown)


src/editor/editor-contract-test.ts
```diff
@@ -406,4 +407,26 @@ console.log("hello");
...
+
+  test("should return the list of code references in same file", async () => {
+    const code = `function add(num1, num2) {
+return num1 + num2;
+}
+add(1, 2);
+`;
+    const editor = await createEditorOn(code);
+    const file =
+      __dirname + "/adapters/vscode-editor-tests/abracadabra-vscode-tests.ts";
```
nicoespeon, line 419: 🍰  Is it necessary to use such a specific file? 
Could we make it relative, like `const filePath = new RelativePath('./code-references.playground.ts')`?

nicoespeon (approved):
👍 Thanks for adding tests. I had 2 comments but no blockers. The tests are passing and are easy to read so that's good for me.

I see there was changes in the InMemory editor implementation, so maybe that means these tests were helpful after all 😃

nicoespeon:
@11joselu I handled the comments, so you may need to update the follow-up PRs to include the latest changes on main 😉

src/editor/editor-contract-test.ts
11joselu: I just saw your changes, the main issue was with TestRunner itself 👍🏽

PR #747, opened 2022-11-05
Change signature II: Handle arrow functions 

11joselu:
Hey :) 

**This PR depends on https://github.com/nicoespeon/abracadabra/pull/746**

I was able to change function signature of an arrow function, but there are some doubts when to trigger "Change signature" action.

In this iteration we shoud put the Cursor on function asignation:

Untitled-2022-02-17-2107

But maybe we want to trigger "Change signature" over variable declaration. 

here-trigger

**Where do you think we should trigger it** @nicoespeon ?

### DIFF
--- a/src/refactorings/change-signature/change-signature.test.ts
+++ b/src/refactorings/change-signature/change-signature.test.ts
@@ -26,15 +26,6 @@ describe("Change Signature", () => {
           }`
       },
       {
-        description: "when there are a arrow function",
-        code: `const item = [cursor](a, b) => {
-            return a + b;
-          }`,
-        expected: `const item = (b, a) => {
-            return a + b;
-          }`
-      },
-      {
         description: "when there are a defined function with references",
         code: `function [cursor]add(a, b) {
             return a + b;
--- a/src/refactorings/change-signature/change-signature.test.ts
+++ b/src/refactorings/change-signature/change-signature.test.ts
@@ -14,7 +14,7 @@ describe("Change Signature", () => {
     code: Code;
     expected: Code;
   }>(
-    "In same file with a function declarations",
+    "In the same file with function declarations",
     [
       {
         description: "when there are a function without references",
@@ -198,10 +198,10 @@ describe("Change Signature", () => {
     code: Code;
     expected: Code;
   }>(
-    "In same file with a arrow function declarations",
+    "In the same file with arrow function declarations",
     [
       {
-        description: "when there are an arrow function without references",
+        description: "when there is an arrow function without references",
         code: `const add = [cursor](a, b) => {
             return a + b;
           }`,
@@ -210,7 +210,7 @@ describe("Change Signature", () => {
           }`
       },
       {
-        description: "when there are an arrow function with references",
+        description: "when there is an arrow function with references",
         code: `const add = [cursor](a, b) => {
             return a + b;
           }
@@ -224,7 +224,7 @@ describe("Change Signature", () => {
       },
       {
         description:
-          "when there are a defined arrow function in multiple lines with references",
+          "when there is a defined arrow function in multiple lines with references",
         code: `const add = [cursor](
           a,
           b) => {
--- a/src/refactorings/change-signature/change-signature.test.ts
+++ b/src/refactorings/change-signature/change-signature.test.ts
@@ -321,6 +321,57 @@ describe("Change Signature", () => {
               }
             ]
           }
+        },
+        {
+          description: "that import arrow function",
+          setup: {
+            currentFile: {
+              code: `export const add = [cursor](a, b) => {
+              return a + b;
+            }`,
+              path: new AbsolutePath("/temp/module.ts")
+            },
+            otherFiles: [
+              {
+                code: `import {add} from './module';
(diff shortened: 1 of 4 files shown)


nicoespeon:
@11joselu I think it's fine to restrict the scope of the trigger to the function params itself for now 👍 
We'll see from usage if it feels like it would be intuitive to have it on the variable assignment too. But I think it's fair to put your cursor on the params if that's what you want to change.

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -14,7 +14,7 @@ describe("Change Signature", () => {
     code: Code;
     expected: Code;
   }>(
-    "In same file",
+    "In same file with a function declarations",
```
nicoespeon, line 17:
```suggestion
    "In the same file with function declarations",
```

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -185,6 +194,67 @@ describe("Change Signature", () => {
     }
   );
 
+  testEach<{
+    code: Code;
+    expected: Code;
+  }>(
+    "In same file with a arrow function declarations",
```
nicoespeon, line 201:
```suggestion
    "In the same file with arrow function declarations",
```

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -25,6 +25,15 @@ describe("Change Signature", () => {
             return a + b;
           }`
       },
+      {
+        description: "when there are a arrow function",
```
nicoespeon, line 29: Shouldn't this test be moved to the new "testEach" that is about arrow functions?

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -185,6 +194,67 @@ describe("Change Signature", () => {
...
   );
 
+  testEach<{
+    code: Code;
+    expected: Code;
+  }>(
+    "In same file with a arrow function declarations",
+    [
+      {
+        description: "when there are an arrow function without references",
```
nicoespeon, line 204:
```suggestion
        description: "when there is an arrow function without references",
```

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -185,6 +194,67 @@ describe("Change Signature", () => {
...
+        code: `const add = [cursor](a, b) => {
+            return a + b;
+          }`,
+        expected: `const add = (b, a) => {
+            return a + b;
+          }`
+      },
+      {
+        description: "when there are an arrow function with references",
```
nicoespeon, line 213:
```suggestion
        description: "when there is an arrow function with references",
```

src/refactorings/change-signature/change-signature.test.ts
```diff
@@ -185,6 +194,67 @@ describe("Change Signature", () => {
...
+          `,
+        expected: `const add = (b, a) => {
+            return a + b;
+          }
+          add(2, 1);
+          `
+      },
+      {
+        description:
+          "when there are a defined arrow function in multiple lines with references",
```
nicoespeon, line 227:
```suggestion
          "when there is a defined arrow function in multiple lines with references",
```

src/refactorings/change-signature/change-signature.ts
```diff
@@ -135,13 +144,32 @@ function updateCode(
...
+      if (!t.isVariableDeclarator(path.parent)) return;
+
+      if (!path.parent.loc) return;
+
+      // For code Line always starts at 1. So we need to start at 0.
+      // Starting at 1 will get body function instead of function declaration
+      const arrowSelection = new Selection(
+        [path.parent.loc.start.line - 1, path.parent.loc.start.column],
+        [path.parent.loc.end.line - 1, path.parent.loc.end.column]
+      );
```
nicoespeon, line 170: 🍰 I think you should be able to use `Selection.fromAST(path.parent.loc)` for that. It handles the line start at 1 and all, so you can drop the comment in this case.

To be confirmed with tests, but that is a common issue that I solved in the past IMO.

nicoespeon:
@11joselu could you update this PR with the latest main? It includes part 1 + a little change in the contract test that causes a conflict.

11joselu:
> @11joselu could you update this PR with the latest main? It includes part 1 + a little change in the contract test that causes a conflict.

Branch updated :)

I will update Class branch too

src/refactorings/change-signature/change-signature.ts
11joselu: Oh! Nice factory.

Used at https://github.com/nicoespeon/abracadabra/commit/d3a78c1001b98783c49aef7234faef6160159ca4
✅ Test OK!

src/refactorings/change-signature/change-signature.test.ts
11joselu: Is a duplicated test. 

Removed at https://github.com/nicoespeon/abracadabra/pull/747/commits/3579f40d0279cb13544161575aabacf263c42ca2

Nice catch!

PR #750, opened 2022-11-10
Add "Flip Yoda Condition" refactoring

### LINKED ISSUE #749: New refactoring: Yoda flip
Some people love [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions), some people hate them. While Yoda conditions do have their [pros and cons](https://knowthecode.io/yoda-conditions-yoda-not-yoda) flipping a condition involves cut'n'paste, some cursor movements and sometimes logic reverse. It may be not an issue for a couple of places in code but when refactoring a big code base this raises the amount of "bleeps" quite significantly. 

`'🙂' === [cursor]emotion` -> `Ctrl+.` -> `Yoda flip` -> `emotion[cursor] === '🙂'`

Refactoring should also account for comparison operator:
- `==`, `===`, `!=`, `!==` should remain unchanged
- `>`, `>=`, `<`, `<=` should be flipped

j4k0xb:
Closes #749 
`+`, `*`, `&`, `|`, `^` can also be flipped, but maybe the name should be changed since they're not conditions.
At least with numbers it behaves the same but I just noticed that `+` changes the order of strings when flipping.. ignore, add type checking or exclude this operator?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **[New Refactoring]** "Change Signature" thanks to @11joselu's great work to find function references across files.
 
+- **[New Refactoring]** "Flip Yoda Condition"
+
 ### Fixed
 
 - "Toggle Braces" and "Convert to ForEach" used to transform the outside-most wrapper that would match the transformation. Now, they transform the selected target, even if it's deeply nested. Props to @j4k0xb for finding this one.
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "onCommand:abracadabra.extractInterface",
     "onCommand:abracadabra.flipIfElse",
     "onCommand:abracadabra.flipTernary",
+    "onCommand:abracadabra.flipYodaCondition",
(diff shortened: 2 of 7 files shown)


j4k0xb:
Yeah makes sense, I've updated it to only apply to conditions now
There isn't that much use for the other ones anyways

nicoespeon:
Thanks for implementing this @j4k0xb, that looks great and the tests are perfect!

I gave it a little more thoughts and I'll make a change before I release this. Indeed, this refactoring seems more generally called "Flip operator". In fact, "Yoda Condition" is the result of using this refactoring to get the value tested against in the first position. It would look weird to have it say "Yoda Condition" for the reverse operation.

Said differently: "Flip Operator" is the refactoring that can achieve the "Yoda Condition", which is the result.

Thus, I think it makes sense to introduce back the `||` and `&&` operators again. I can see it helpful to be able to transform:

```js
if (isValid || canTransform) 
```

into:

```js
if (canTransform || isValid) 
```

I'll let the other ones like `+` on the side for now.

I'm gonna merge this and do the changes. Thanks!

nicoespeon:
@all-contributors please add j4k0xb for code

PR #772, opened 2022-12-04
Do not show 'Change signature' quick fix when a function/method does not has parameters

11joselu:
Hide "Change signature" for function/method without parameters

Error:
image
image

New:

image
image

Even for React Components

Error:
image

New
image

### DIFF
--- a/src/refactorings/change-signature/change-signature.test.ts
+++ b/src/refactorings/change-signature/change-signature.test.ts
@@ -1,8 +1,10 @@
+import * as t from "../../ast";
 import { InMemoryEditor } from "../../editor/adapters/in-memory-editor";
 import { AbsolutePath, Code, ErrorReason } from "../../editor/editor";
+import { Selection } from "../../editor/selection";
 import { testEach } from "../../tests-helpers";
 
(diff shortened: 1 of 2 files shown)


nicoespeon:
Thanks for the fine-tuning 👍 

Could you add some tests to ensure that it doesn't get broken in the future? I think you could replicate what was done here for instance: https://github.com/nicoespeon/abracadabra/blob/303ec414d5ab47b44e945088febf9769d5320d46/src/refactorings/convert-to-template-literal/convert-to-template-literal.test.ts#L12-L41

11joselu:
> Thanks for the fine-tuning +1
> 
> Could you add some tests to ensure that it doesn't get broken in the future? I think you could replicate what was done here for instance:
> 
> https://github.com/nicoespeon/abracadabra/blob/303ec414d5ab47b44e945088febf9769d5320d46/src/refactorings/convert-to-template-literal/convert-to-template-literal.test.ts#L12-L41

I have added some tests https://github.com/nicoespeon/abracadabra/pull/772/commits/0994c557ec29631fe2084da2c4a493f935e03b0e

PR #797, opened 2023-01-25
[New Feature] Persisting Highlights

nicoespeon:
Besides refactorings, Abracadabra now also gives you some extra features that are here to support refactoring work.

### Highlight Identifiers

toggle-highlights

When working with Legacy Code, a very useful refactoring consists in extracting pure logic out of the I/O code that is hard to test (eg. HTTP calls).

To help you spot all the I/O bits in a chunk of code, Abracadabra can highlight them for you!

| Feature               | Keybinding         | On Mac  |
| :-------------------- | :----------------- | :------ |
| Toggle Highlight      | `Ctrl + H`         | `⌃ H`   |
| Remove All Highlights | `Ctrl + Shift + H` | `⌃ ⇧ H` |

1. Put your cursor over an Identifier you want to highlight
2. Press `Ctrl + H`

All references to this Identifier will be highlighted. The highlight will persist even when you modify the code.

To remove, either:

- Put your cursor over a highlighted reference and press `Ctrl + H` again
- Press `Ctrl + Shift + H` from anywhere, to remove all highlights

This is handy for marking I/O code when refactoring, but you can use it to highlight any interesting Identifier and its references. Here are a few typical use cases:

- Mark all usages of a variable, to help you regroup them
- Identify variables that are always passed together in function calls

## Remaining TODOs

- [ ] Support "Extract & Destructure" scenario (it still messes up the highlights)
- [ ] Handle when identifying a `let cost` and usage within if/else

There will likely be a bunch of edge-cases that won't work to start with. That's fine. I'll fix them as I come across them, or when they are reported.


nicoespeon:
Merging this now. I'll handle the edge-cases with follow-ups.

Issue #820, opened 2023-02-27
change activationEvent to onLanguage


tjx666:
## Is this request related to a problem? Please describe.

When we not open the js related language, this extension needn't activated.
Because many extension choose activated at `onStartupFinished`, that will slow down each extensions startup speed ref: https://github.com/microsoft/vscode/issues/172933#issuecomment-1411119225

https://code.visualstudio.com/api/references/activation-events#onLanguage

## Describe the solution you'd like

```json
"activationEvents": [
    "onLanguage:javascript",
    "onLanguage:javascriptreact",
    "onLanguage:typescript",
    "onLanguage:typescriptreact",
    "onLanguage:vue",
    "onLanguage:svelte",
]
```

## Additional context

_Add any other information or screenshots about the feature request here (e.g. a gif showing how another editor does it)._

PR #830, opened 2023-03-05
perf: optimize activationEvents

tjx666:
#820

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,12 @@
     "@babel/types": "7.20.7"
   },
   "activationEvents": [
-    "onStartupFinished"
+    "onLanguage:javascript",
+    "onLanguage:javascriptreact",
+    "onLanguage:typescript",
+    "onLanguage:typescriptreact",
+    "onLanguage:vue",
+    "onLanguage:svelte"
   ],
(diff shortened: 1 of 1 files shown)


tjx666:
I find the souce code is not very many, but the project typescript intellisence is very slow, don't konw why.

nicoespeon:
Interesting. So that means reverting the change that I thought would improve the situation: https://github.com/nicoespeon/abracadabra/commit/1cfac73bf8de14edb77cdb8cdd523d7d1b37cd5a

My understanding was that "onStartupFinished" would indeed trigger every time but won't delay the editor startup vs. "onLanguage" could delay the startup if a JS/TS file was already open in the project. Did I understand that wrong and actually made the situation worse?

tjx666:
One problem onStartupFinished cause this extension will be actived every time the vscode startup even you open a empty workspace. Because you run all the modules synchronously

nicoespeon:
Got it, thanks for the details. Let's go with this improvement. Further improvements would be to go for the lazy loading, as you mentioned. Or we should also measure what takes so long and optimize that.

But this is acceptable to me so far.

PR #843, opened 2023-03-24
Add `Wrap in JSX fragment` refactoring

### LINKED ISSUE #791: Wrap with JSX Fragment

Adding a React Fragment around a component because you want to add a sibling to it is quite tedious.

When selecting a JSX tag you could be offered a refactoring option to wrap this tag in a React Fragment (`<>...</>`). If possible also when selecting multiple tags.

## Additional context

/

---

Thanks for the project, very helpful :)

DerTimonius:
Hi everyone!
Thanks for this awesome extension, just the right amount of magic a developer needs 🪄 

This PR will, once finished, add a refactoring that wraps the selection in a `JSX fragment` if it is a JSX element.
Before:
```jsx
return <div>Something witty goes here</div>;
```
After:
```jsx
return (
  <>
    <div>Something witty goes here</div>
  </>
);
```

### Still Todo:
- [ ] write tests
- [ ] adding the parentheses not yet working as intended

Closes #791

### DIFF
--- a/package.json
+++ b/package.json
@@ -344,6 +344,11 @@
         "command": "abracadabra.removeAllHighlights",
         "title": "Remove All Highlights",
         "category": "Abracadabra"
+      },
+      {
+        "command": "abracadabra.wrapInJsxFragment",
+        "title": "Wrap in JSX Fragment",
+        "category": "Abracadabra"
       }
     ],
     "keybindings": [
@@ -1464,6 +1469,14 @@
         {
           "command": "abracadabra.toggleBraces",
           "when": "editorLangId == svelte"
+        },
+        {
+          "command": "abracadabra.wrapInJSXFragment",
+          "when": "editorLangId == javascriptreact"
+        },
+        {
+          "command": "abracadabra.wrapInJSXFragment",
+          "when": "editorLangId == typescriptreact"
         }
       ]
     }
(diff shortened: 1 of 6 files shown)


DerTimonius:
Sorry for taking so long, it has been quite a busy week.
I was able to correctly add the parentheses for single line return statements and also to add the fragments in nested return statements, but I still have two problems:
1. If the return statement is parenthesized, the JSX fragments are added outside, not inside the parentheses.
2. If you want to add the fragments to the nested return statements, you also add them to the parent return statement.

Do you maybe give me some clues on what I'm missing? 🙂 (first time working with AST, a lot to dive into!)

DerTimonius:
Thanks for your help!
I just added another test for a nested return statement to check if it's working as intended and it does :+1:

nicoespeon:
@all-contributors please add DerTimonius for code

nicoespeon:
Thank you for tackling this @DerTimonius 🙏 

If you are motivated, I thought it would be cool to have the reverse refactoring: "Remove JSX fragment"
It should be easy, but there is a gotcha: it should only perform the refactoring if the fragment can indeed be removed. So this fragment can't be removed for instance:

```tsx
return <>
  <SomeComponent />
  <AnotherComponent />
</>;
```

DerTimonius:
That's a great idea, I would love to work on that!

Thanks again for your help @nicoespeon

PR #864, opened 2023-04-15
Add refactoring to remove a JSX fragment

DerTimonius:
This PR will add a new refactoring that removes a JSX fragment, if there is just one child component inside that fragment. It's pretty much the opposite from my previous PR #843 
The refactoring will not work if there are multiple children to the JSX fragment.

I also added two tests with working examples and also a failing one in the `"should show an error message if refactoring can't be made"` test.


DerTimonius:
Hey @nicoespeon is there anything more that should be added?

nicoespeon:
Totally understand. I should have let you know that earlier. It's just temporary lousy timing on my side with a few deadlines looking upon my shoulder. But I'm not forgetting you 😉

nicoespeon:
And I finally found time to complete my review and document this new refactoring. I was thinking about exposing both behind a "Toggle JSX Fragment", but after I played with this I don't feel it's really necessary. Anyway, no reason to delay this further more.

Thank you for your contribution @DerTimonius 👏

DerTimonius:
Thanks for the review and merge!

PR #931, opened 2023-10-21
Upgrade dependencies

nicoespeon:
It's been a while. Make sure CI doesn't break when everything is upgraded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@babel/traverse": "7.20.12",
     "@babel/types": "7.20.7",
     "@eslint/eslintrc": "1.3.3",
-    "@typescript-eslint/parser": "5.50.0",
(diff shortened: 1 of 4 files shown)


nicoespeon:
It's a mess aha 😅 

Turns out upgrading TS creates incompatibilities with at least `@typescript-eslint/parser` that we use for "Extract useCallback()" refactoring. They require changing the module resolution of TS (https://github.com/typescript-eslint/typescript-eslint/issues/7284) but that creates a bunch of new problems with the way we run the contract tests (code is compiled and run in a VS Code environment) as well as the web build.

I don't want to spend too much time on this for now. I suspect the simplest way to move forward would be to drop the React-specific refactoring "Extract useCallback". We may eventually get there if this is blocking us on an old version of TS.

Issue #416, opened 2021-10-05
Destructure Object is too slow


nicoespeon:
See https://github.com/nicoespeon/abracadabra/issues/415 for context

We had to disable the Action Provider for this refactoring because it would cause VS Code to hang.

The thing is:
- creating a new TS program every time the refactoring runs is expensive
- because of the Action Provider behavior, it would run on every click/keypress

Solutions to fix:
1. Don't create a new TS program on every Action Provider run (similar optimization we did for parsing the AST only once)
2. Probably better => don't create a new TS program at all but tap into VS Code's one instead

Going with 2) will also unlock multi-files refactoring. Definitely the path I want to explore.

- [ ] Fix the perfs issue
- [ ] Tell OliverJAsh about findings on Language Service Plugins
- [ ] Enable the Action Provider again
- [ ] Update REFACTORING.md to tell it's now available as a Quick Fix

## QA

[Monitor performance when running the extension](https://www.nicoespeon.com/en/2019/11/fix-vscode-extension-performance-issue/) on a large TS file. This snippet should do:

<details><summary>This snippet should do:</summary>

```tsx
interface GitgraphOptions {
  template?: TemplateName | Template;
  orientation?: Orientation;
  reverseArrow?: boolean;
  initCommitOffsetX?: number;
  initCommitOffsetY?: number;
  mode?: Mode;
  author?: string;
  branchLabelOnEveryCommit?: boolean;
  commitMessage?: string;
  generateCommitHash?: () => Commit["hash"];
  compareBranchesOrder?: CompareBranchesOrder;
}

interface RenderedData<TNode> {
  commits: Array<Commit<TNode>>;
  branchesPaths: BranchesPaths<TNode>;
  commitMessagesX: number;
}

class GitgraphCore<TNode = SVGElement> {
  public orientation?: Orientation;
  public get isHorizontal(): boolean {
    return (
      this.orientation === Orientation.Horizontal ||
      this.orientation === Orientation.HorizontalReverse
    );
  }
  public get isVertical(): boolean {
    return !this.isHorizontal;
  }
  public get isReverse(): boolean {
    return (
      this.orientation === Orientation.HorizontalReverse ||
      this.orientation === Orientation.VerticalReverse
    );
  }
  public get shouldDisplayCommitMessage(): boolean {
    return !this.isHorizontal && this.mode !== Mode.Compact;
  }

  public reverseArrow: boolean;
  public initCommitOffsetX: number;
[...]

PR #934, opened 2023-10-28
Remove costly refactorings

nicoespeon:
This PR removes 2 refactorings:

1. Destructure Object
2. Extract useCallback()

## Destructure Object

This one was not working well already (#416). It depended on a POC to resolve the types at runtime, which required the inclusion of the whole "typescript" library in the bundle.

Furthermore, I realized that TS 3.4 introduced a similar refactoring to VS Code: https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/#convert-parameters-to-destructured-object

## Extract useCallback()

This refactoring was specific to React. In short, it was applying the auto-fix of the [`eslint-plugin-react-hooks`](https://www.npmjs.com/package/eslint-plugin-react-hooks) package.

While this was useful, it's an edge-case that only applies to React codebases. It was not really causing performance issues (although any extra refactoring has a cost), but it required to include many packages in the bundle, including the "typescript" library (used by ESLint and the plugin that was also included).

## Impact

The size of the extension bundle went down from **~9MB to ~1.5MB**:

| Before | After |
|---|---|
| before | after |

And for the browser bundle (the one that runs on VS Code Web), it went down from **~5.5MB to ~1.6MB**:

| Before | After |
|---|---|
| before-browser | after-browser |


nicoespeon:
And somehow it seems that Travis stopped working… We'll see, but eventually, we will move to GitHub Actions (which we already use for deployments), so that would be one less moving part.

Issue #902, opened 2023-07-18
Abracadabra extension causes performance issues with other extensions (prettier formatting, etc.)


dinofx:
I have prettier set as my formatter and format on save enabled.  If I:

1. open a large JS file, for example a bundle created by webpack with 50,000 lines, then
2. open a file with 15 lines of code
3. Add a few extra newlines to the end of the small file and save

I see a popup and the save is delayed for a long time.
![image](https://github.com/nicoespeon/abracadabra/assets/7581987/d63bbc72-13b8-423d-87ef-73ed8fc1a1ae)

If I disable this extension, the problem goes away.  I tried setting an ignore pattern to include the folder containing the large JS file, but I wasn't able to workaround the delay.

The problem has affected other extensions as well.  For example, I tried several extensions that simply display the absolute character offset.  When working in large files, these extensions would all display an initial offset but then would stop updating as the caret moved around.  Also I couldn't invoke the "go to offset" command for the extensions.

maxxxxxdlp:
Seeing very bad performance when doing "Go to definition" in a large typescript definitions file (115k lines):

https://github.com/nicoespeon/abracadabra/assets/40512816/e741a4d0-a33d-4e33-a53d-2973b4c986bd

Using VS Code's [extension bisect](https://code.visualstudio.com/blogs/2021/02/16/extension-bisect), found that Abracadabra is the cause - when it is disabled, performance in the 115k line file is awesome

nicoespeon:
Thanks for reporting. To be honest, I'm not surprised. I didn't have perfs in mind when I started implementing these refactorings. 

It was working fine on files of reasonable size at first. The more refactorings we added to the catalog, the worse the perfs went, especially on large files. The exclusion feature was supposed to help, but it doesn't seem to.

We have had other obvious instances of this (see https://github.com/nicoespeon/abracadabra/issues/416) and I believe something like https://github.com/nicoespeon/abracadabra/issues/836 would be the solution. I haven't taken the time to dig into it, but I will think about it. In the meantime, https://marketplace.visualstudio.com/items?itemName=p42ai.refactor may be a better alternative so you have a catalog of refactorings, but with better perfs (I think).

nicoespeon:
@dinofx that's a great idea!

[...]

PR #981, opened 2024-01-03
Fix action provider file ignore checks

j4k0xb:
Trying to execute a refactoring quick action in an untitled tab used to throw an error:

```
[error] EntryNotFound (FileSystemError): Error: ENOENT: no such file or directory, open '/Untitled-1'

at Function.e (/opt/visual-studio-code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:143:26791)
at Object.readFile (/opt/visual-studio-code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:143:25177)
at RefactoringActionProvider.isNavigatingAnIgnoredFile (/tmp/abracadabra/src/action-providers.ts:48:25)
at RefactoringActionProvider.provideCodeActions (/tmp/abracadabra/src/action-providers.ts:23:9)
at N.provideCodeActions (/opt/visual-studio-code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:143:44632)
```

Now it uses the text/lineCount from the document instead of reading from the fs:
image
Regarding the line count vs file size (#902) I've kept it the same for now to avoid a breaking change but it's up to you 😅

And while testing I also noticed that `node_modules` in the workspace root isn't ignored because the relative path doesn't start with `/`
Would have written tests for it but not sure where or if they're necessary for this

### DIFF
--- a/src/action-providers.ts
+++ b/src/action-providers.ts
@@ -20,7 +20,7 @@ export class RefactoringActionProvider implements vscode.CodeActionProvider {
   async provideCodeActions(document: vscode.TextDocument) {
     const NO_ACTION: vscode.CodeAction[] = [];
 
-    if (await this.isNavigatingAnIgnoredFile(document.uri.path)) {
+    if (this.isNavigatingAnIgnoredFile(document)) {
       return NO_ACTION;
     }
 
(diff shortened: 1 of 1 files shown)


nicoespeon:
Awesome @j4k0xb 🙌 

It's looking good for me. It's way better now that we don't have to make an async check for the file length 😉
I will look into using the file size instead, as we discussed.

PR #1038, opened 2024-07-06
Fix inline variable with shadowed parameter

### LINKED ISSUE #1037: Inline variable overwrites unrelated parameter
## Describe the bug

Inlining a variable doesn't only replace the references, but also parameters that shadow it.
Seems to only happen for methods/functions/setters inside an object.

## How to reproduce

Move the cursor to the first foo and execute `Inline variable`.

```js
var foo = 1;
console.log(foo);
fn({
  set(foo) {
    return foo;
  },
});

```

## Expected behavior

Only inline references of the variable.

```js
console.log(1);
fn({
  set(foo) {
    return foo;
  },
});

```

## Current behavior

It also replaces the `foo` parameter.

```js
console.log(1);
fn({
  set(1) {
    return 1;
  },
});
```

## Additional information

- Version of the extension impacted: _v9.3.0_
- Related: #316

j4k0xb:
Closes #1037

### DIFF
--- a/src/ast/scope.ts
+++ b/src/ast/scope.ts
@@ -2,10 +2,7 @@ import { Binding, NodePath } from "@babel/traverse";
 import * as t from "@babel/types";
 import { first } from "../array";
 import { getImportDeclarations, TypeDeclaration } from "./domain";
-import {
-  areEquivalent,
-  isFunctionDeclarationOrArrowFunction
-} from "./identity";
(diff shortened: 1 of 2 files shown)


nicoespeon:
Neat, and thanks for the test!

nicoespeon:
FYI, this is being released with 9.4.0

PR #1142, opened 2025-04-08
Don't make refactoring asynchronous as default

zoodogood:
The generator creates a function with a return value of Promise type. The expected return type of the function is EditorCommand.  A mismatch violates the [executeRefactoring function strategy](https://github.com/nicoespeon/abracadabra/blob/9e5f9044c17ad206a1312e7e78a1b8988322986a/src/refactorings.ts#L143); and the returned error message is implicit for quick correction

I need a time to find a source of this error


nicoespeon:
Good catch @zoodogood, thank you!

Issue #1175, opened 2025-06-06
[New Refactoring] Extract Parameter


nicoespeon:
## Is this request related to a problem? Please describe.

This refactoring is to experiment with contributing to Abracadabra by implementing a new move. It's also a good way to add a new transformation.

## Describe the solution you'd like

Considering a VariableDeclaration within a FunctionDeclaration, we want to be able to transform the VariableDeclaration into an AssignmentDeclaration in the params of the FunctionDeclaration.

Concretely, if we have:

```js
function doSomething(myVar) {
  const anotherVar = "some value";

  // …
}
```

We want to be able to refactor it into:

```js
function doSomething(myVar, anotherVar = "some value") {
  // …
}
```

And we should handle all the different scenarios and edge-cases we can think about.

## Additional context

This refactoring was suggested at [Alpes Craft 2025](https://www.alpescraft.fr/). We started working on it in a mob programming session with the participants.

PR #1177, opened 2025-06-07
[New Refactoring] Extract Parameter

nicoespeon:
Implements #1175 

Currently in progress. This is the result of our mob programming session at Alpes Craft 2025. 

**Feel free to take it over if you want**. Just mention you're working on it. If no-one takes it by mid-July, I will eventually do and finish the implementation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -211,6 +211,11 @@
         "category": "Abracadabra"
       },
       {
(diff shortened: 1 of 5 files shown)


nicoespeon:
Implemented with #1192

PR #1192, opened 2025-07-13
Implement first POC in mob at Alpes Craft

pinage404:
Implements https://github.com/nicoespeon/abracadabra/issues/1175

Superseds #1177 

Do you see any missing cases ?

I didn't find a way to remove the `@ts-expect-error`

### DIFF
--- a/package.json
+++ b/package.json
@@ -211,6 +211,11 @@
         "category": "Abracadabra"
       },
       {
+        "command": "abracadabra.extractParameter",
+        "title": "Extract Parameter",
+        "category": "Abracadabra"
+      },
+      {
         "command": "abracadabra.flipIfElse",
         "title": "Flip If/Else",
         "category": "Abracadabra"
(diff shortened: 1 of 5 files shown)


nicoespeon:
@pinage404 it's all good!

I managed to fix the type error by diffing the 2 types and adding guards for the types that could not be supported. I put that logic in the `createVisitor()` since the `updateCode()` logic should just transform the code and not filter out candidates (otherwise we may suggest refactorings that can't actually be executed).

I've added a few more tests and scenarios I could think about. The feature is looking good enough now. I'm gonna merge this and document the feature—not sure if I'm releasing it right away or wait a bit to gather more changes before bumping the version, but it will soon be part of Abracadabra.

Thank you again for your help on this, really appreciated 🙏

nicoespeon:
@all-contributors please add pinage404 for code

PR #1195, opened 2025-07-15
refactor DRY is function check

pinage404:
Following #1192

A small refactor to check at only one place if it's inside a function

I let you merge or close this PR if you don't want it, it's totally fine =)


nicoespeon:
Good stuff, I like the refactoring and the extra test 👍

